PR #17, opened 2021-04-01
fix broken link

pfeilbr:
Perm fix might be rename this dir (`kinesis-lambda-efo`) to (`kinesis-lambda-enhanced-fan-out`).  Not sure if this is generated via automation


jbesw:
Thanks for catching this!

PR #26, opened 2021-04-05
fix: removing unnecessary roles

ljacobsson:
*Description of changes:*
Removed `AWS::IAM::Role` resource and made use of the `Policies` array on both the `AWS::Serverless::Function` and `AWS::Serverless::StateMachine`

### DIFF
--- a/lambda-sfn/template.yaml
+++ b/lambda-sfn/template.yaml
@@ -13,7 +13,9 @@ Resources:
       Handler: app.handler
       Runtime: nodejs12.x
       Timeout: 3
-      Role: !GetAtt LambdaExecRole.Arn
+      Policies:
+        - StepFunctionsExecutionPolicy:
+            StateMachineName: !GetAtt StateMachine.StateMachineName
       Environment:
         Variables:
           StateMachineArn: !GetAtt StateMachine.Arn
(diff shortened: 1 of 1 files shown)


gadavis2:
If you remove an IAM Role from a template, then any reference to that role would need to be updated/removed as well. However, I intentionally included a role in the apigw-http-api-lambda/template.yml for demonstration purposes. Many real-world situations will require the function to have more than just the basic execution role policy or the managed policies provided by SAM.

ljacobsson:
Not sure what's considered best practice, but I think one of the great things about SAM is that the creation of the execution roles. I haven't come across any scenario where one is needed apart from edge cases where it has made sense to have one shared role for many functions.

The `Policies` array accepts managed roles, SAM policy templates and custom composed policies:
```
MyFunction:
  Type: AWS::Serverless::Function
  Properties:
    Policies:
      - AmazonSQSFullAccess # <-- bad
      - SQSSendMessagePolicy:
          QueueName: !GetAtt MySqsQueue.QueueName
      - Version: 2012-10-17
        Statement:
          - Sid: Statement1
            Effect: Allow
            Action:
              - 'sqs:SendMessage'
              - 'sqs:SendMessageBatch'
            Resource:
              - !GetAtt MySqsQueue.Arn
  [...]
```

PR #106, opened 2021-08-24
Update event bus name in sample event

jhart0:
Update sample event to use custom EventBusName specified in SAM template. Remove duplicate Detail property. 


jbesw:
Thanks Josh!

PR #199, opened 2021-11-09
Improvement/lambda-dynamodb-cdk

GlennChia:
*Issue #, if available:*

*Description of changes:*

Minor improvements to the pattern

- [fix] Add RemovalPolicy.DESTROY to the DynamoDB table because I realised that it wasn't destroyed on cdk destroy
- [documentation]: add note on AWS CLI v2 requirement for test case. Add example for test case


jbesw:
Thanks for this update!

PR #221, opened 2021-11-25
Fix broken rule

SimonCMoore:
It seems the log group arn has a trailing :* which deploys fine without errors but did not work and the event never reached the eventbus. Maybe there is a better workaround but constructing the arn without the * seems to resolve it. 

Related to the extra star in console clicking from the rule through to the target shows error message in the cloudwatch console "An error occurred while describing log groups.
1 validation error detected: Value '/aws/events/sfn-eventbridge:*' at 'logGroupNamePrefix' failed to satisfy constraint: Member must satisfy regular expression pattern: [\.\-_/#A-Za-z0-9]+"

### DIFF
--- a/sfn-eventbridge/template.yaml
+++ b/sfn-eventbridge/template.yaml
@@ -62,7 +62,7 @@ Resources:
           - prefix: ""
       State: ENABLED
       Targets:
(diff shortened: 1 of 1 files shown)


jbesw:
Thanks for the improvement!

PR #228, opened 2021-11-30
New serverless pattern - ta-eventbridge-lambda-s3

seandall:
*Issue #, if available:*

*Description of changes:*
This Serverless pattern shows how you can automate actions from a trusted advisor trigger. Services used in this pattern include Trusted Advisor, EventBridge, Lambda, and SNS


jbesw:
Sean, can you verify this PR? It has over 1700 files - it should be no more than 5-20.

PR #370, opened 2022-02-09
Added missing app.js in SQS-Lambda using CDK

praghu1:
*Issue #, if available:*

*Description of changes:*

Added missing `app.js` in SQS-Lambda using CDK


jbesw:
Thanks for the improvement!

PR #393, opened 2022-02-14
sqs-lambda-rust: change to arm64 architecture

ymwjbxxq:
*Issue #, if available:*

*Description of changes:*
Because the Makefile was already using the cross build I think is best to use arm64 architecture as default.
To be able to use arm64 architecture, we must install [CROSS](https://github.com/rust-embedded/cross)


jbesw:
Thanks for the improvement!

PR #449, opened 2022-02-25
Avoid buffering image conversion with streams

lmammino:
*Issue #, if available:*

The illustrated solution buffers the thumbnail entirely in memory before starting to send data back to the user.

*Description of changes:*

The proposed change keeps sharp operating in streaming mode and since `writeGetObjectResponse` supports a readable stream in the `Body` property, this approach should start to stream bytes back to the user while the thumbnail is generated.


jbesw:
Thanks for the improvement!

PR #461, opened 2022-03-08
Lambda sqs sls - Serverless Framework

extremenelson:
*Issue #, if available:*

*Description of changes:*
Added in example of Lambda sending messages to a SQS queue using the Serverless Framework


jbesw:
Hello! We currently don't support Serverless Framework but I hope to add this down the road. I'll close for now but reopen when we have the capacity to support it.

jbesw:
Thanks for the contribution! This will be live later today at serverlessland.com/patterns/lambda-sqs-python-sls.

PR #463, opened 2022-03-10
feature: add dynamodb streams to lambda function pattern via terraform

eelayoubi:
Deploy DynamoDB table with stream enabled, Lambda function that process the stream records, via Terraform


jbesw:
Thanks for your contribution! This is now live at https://serverlessland.com/patterns/dynamodb-lambda-filters-terraform.

PR #466, opened 2022-03-11
feature: add S3 objectCreated event  to lambda function pattern via t…

eelayoubi:
Deploy S3 bucket, with OjectCreated event notification,  Lambda function that processes the new object, via Terraform


jbesw:
Thanks for the contribution! This is now live at https://serverlessland.com/patterns/s3-lambda-terraform.

PR #544, opened 2022-03-16
New Pattern: CloudFront failover with two apigw http api lambda rust in different regions

ymwjbxxq:
*Issue #, if available:*

*Description of changes:*
This pattern follows [Optimizing high availability with CloudFront origin failover](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/high_availability_origin_failover.html)

It will deploy 2 APIGW and 2 Lambda in two different regions, and it will setup a CloudFront distribution on the front with two origins.


jbesw:
Thanks for the contribution! This is now live at https://serverlessland.com/patterns/cloudfront-apigw-failover.

Issue #566, opened 2022-03-28
Add Serverless Framework versions for the most popular patterns


DmitryGulin:
Have a subset of existing patterns implemented with [Serverless Framework](https://www.serverless.com/)

PR #577, opened 2022-03-28
APIGW SQS Lambda - Serverless Framework

DmitryGulin:
#566 

Add support for APIGW (HTTP API) - SQS - Lambda pattern, implemented with Serverless Framework


jbesw:
Thanks! This will be live later today at https://serverlessland.com/patterns/apigw-http-sqs-lambda-sls.

PR #586, opened 2022-03-31
New pattern (Terraform & Python): Amazon FIFO SNS to Amazon FIFO SQS to AWS Lambda to Amazon Kinesis Firehose to Amazon S3 

iobreaker:
*Description of changes:*

The Terraform template in this pattern allow to connect FIFO sqs queue to firehose in order to precess (transform, compress,...) messages sent by a FIFO sns topic and save the result to an S3 bucket using a Lambda function as a bridge to firehose.

This pattern provide a solution to connect SNS Topics and/or SQS queues which is FIFO one to Firehose


jbesw:
Thanks! This is now live at https://serverlessland.com/patterns/sns-sqs-lambda-firehose-s3-terraform.

PR #626, opened 2022-04-17
New serverless pattern: Lambda partial batch response for SQS as an event source with AWS SAM and Rust

ymwjbxxq:
*Issue #, if available:*

*Description of changes:*

- AWS Lambda now supports partial batch response for SQS as an event source. 
- This example shows how to handle a batch of SQS messages in parallel
- This example shows how to handle a partial batch failure 


jbesw:
Thanks for the contribution! This will be live at serverlessland.com/patterns/sqs-lambda-partial-batch-rust-sam shortly.

PR #763, opened 2022-07-23
remove node.js to match other patterns

WayneGreeley:
*Issue #, if available:*

*Description of changes:*

Removing the Node.js from language in example-pattern.json to match other patterns.


jbesw:
Thanks for the update. This will be live shortly.

PR #823, opened 2022-08-23
fixed template issue

srajaw:
*Issue #, if available:*

*Description of changes: this pattern was not working because template had indentation issue and was returning the below error
-------
Resource with id [SESSendFunction] is invalid. Policy at index 0 in the 'Policies' property is not valid
-------

Fixed the template and also update the nodejs version to latest i.e 16x because nodejs12.x is getting deprecated. 


srajaw:
There are 2 updates in this:
1. There was an issue with the template file of the lambda-ses serverless pattern. Also nodejs 12 was being used in it so updated it to Nodejs16.
2. Added a new pattern for step function to invoke cross account lambda function.

jbesw:
For us to merge this, please remove the files in the step-function-lambda-function-crossaccount-terraform directory. If you would like to submit that pattern, please make a separate request.

jbesw:
Thanks for the improvement!

PR #848, opened 2022-09-07
Existing Pattern : Lambda-Glue-S3

Harshi719:
*Issue #, if available:*

*Description of changes:* updated the code to limit the in line permissions to a specific glue job


lambda-glue-s3/app.py
```diff
@@ -68,9 +68,11 @@ def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
             }
         )
         # Add InLine Policy to Lambda to provide access to start the jobs
+        accountID = Stack.of(self).account
+        region = Stack.of(self).region
         customPolicy = iam.PolicyStatement(
             actions=['glue:StartJobRun'],
-            resources=['*']
+            resources=[f"arn:aws:glue:{region}:{accountID}:job/{job.name}"]
```
boyney123, line 75: Hey @Harshi719 

Don't know python that much, but is that `f` character meant to be here?

Is this meant to be 

`resources=["arn:aws:glue:{region}:{accountID}:job/{job.name}"]`?

PR #860, opened 2022-09-15
New serverless pattern - Step function to Step function with CDK

pubudusj:
*Issue #, if available:*
N/A

*Description of changes:*
Example added for starting a Step Function execution as a step in another Step Function with CDK.

### DIFF
--- a/sfn-to-sfn-cdk/lib/sfn-to-sfn-cdk-stack.ts
+++ b/sfn-to-sfn-cdk/lib/sfn-to-sfn-cdk-stack.ts
@@ -38,7 +38,7 @@ export class SfnToSfnCdkStack extends Stack {
         statements: [
           new iam.PolicyStatement({
             actions: ['states:SendTaskSuccess'],
-            resources: ['*'],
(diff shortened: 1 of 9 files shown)


sfn-to-sfn-cdk/lib/sfn-to-sfn-cdk-stack.ts
```diff
@@ -0,0 +1,78 @@
...
+    const childStateMachine = new sfn.StateMachine(this, 'ChildStateMachine', {
+      definition: wait.next(sendTaskTokenSuccess)
+    });
+
+    childStateMachine.role?.attachInlinePolicy(
+      new iam.Policy(this, 'sendTaskSuccessPermission', {
+        statements: [
+          new iam.PolicyStatement({
+            actions: ['states:SendTaskSuccess'],
+            resources: ['*'],
```
boyney123, line 41: @pubudusj maybe we should specify which resources here rather than everything? What you think?

pubudusj: @boyney123 Yeah, agree. This is fixed.

boyney123:
Thanks you @pubudusj I will get this merged, thanks for the contribution and requested changes made :)

PR #887, opened 2022-10-15
New serverless pattern - step-function-sync-rest-api

higoyal01:
*Issue #, if available:*

*Description of changes:* The Serverless Pattern deploys an Amazon API Gateway REST API endpoint along with an Express State machine. This setup illustrate how we can invoke an Express State machine and get the output back from the execution to the client (within 29 seconds) using a Rest API. 

### DIFF
--- a/step-function-sync-rest-api/template.yml
+++ b/step-function-sync-rest-api/template.yml
@@ -1,0 +1,88 @@
+AWSTemplateFormatVersion: '2010-09-09'
+Transform: AWS::Serverless-2016-10-31
+Description: >
+  A Step Functions Express Workflow triggered synchronously by an API Gateway HTTP APIs integration
+
+Resources:
+##########################################################################
+#   STEP FUNCTION                                                        #
+##########################################################################
+
+  StateMachineExpressSync:
+    Type: AWS::Serverless::StateMachine # More info about State Machine Resource: https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-resource-statemachine.html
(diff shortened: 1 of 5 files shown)


step-function-sync-rest-api/template.yml
```diff
@@ -0,0 +1,88 @@
...
+      Name: StateMachineExpressSync
+      DefinitionUri: statemachine/stateMachine.asl.json
+      Policies:
+        -  Version: "2012-10-17"
+           Statement:
+             - Effect: Allow
+               Action:
+                 - "cloudwatch:*"
+                 - "logs:*"
+               Resource: "*"
```
boyney123, line 23: Any chance we could be specific about the resource that is required here, rather than everything?

higoyal01: Scoping down for the resources field for Cloudwatch in Step Functions is not supported, the resources field has to be a wildcard value. This is an existing feature request.

Further, few API actions required for logging do not support specifying resource-level permissions and needs to specify all resources ("*") in the Resource element of your policy statement. For more details on resource restrictions on actions defined by Amazon CloudWatch Logs, please refer:
https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazoncloudwatchlogs.html#amazoncloudwatchlogs-actions-as-permissions

boyney123: Hi @higoyal01 , thanks for the detail and explanation! Will continue to review the pattern

PR #909, opened 2022-11-01
New serverless pattern - EventBridge -> AWS Step Functions -> SQS

shansv:
*Issue #, if available:*

*Description of changes:*
This pattern creates an EventBridge event bus, a Step Functions workflow, and subscribes the Step Functions workflow to be executed when a `customEvent` event is published to the event bus. The Step Functions workflow executes Lambda function for business worflow. AWS Lambda can occasionally experience transient service errors. In this case, invoking Lambda results in a 500 error, such as ServiceException, AWSLambdaException, or SdkClientException. As a best practice, proactively handle these exceptions in your state machine to Retry invoking your Lambda function, or to Catch the error and move to Failed Queue for operational needs. 


shansv:
Corrected the language in the "example-pattern.json"

boyney123:
Thanks @shansv . LGTM!

PR #922, opened 2022-11-09
New serverless pattern - lambda function to lambda function(async) through S3 bucket

SeoyeonPark:
Description of changes:

Lambda function that call another Lambda function asynchronously. 

When the payload size is exceeded Lambda invocation payload(256kb in case of async invocation), then first Lambda function put the payload as an object file to S3 bucket. Then second Lambda function will be triggered by 's3:ObjectCreated' event, and then read the payload object from S3 bucket.

This pattern is helpful when you have to handle large size of data that exceed the limitation payload quota of Amazon SNS, Amazon SQS, Amazon MSK, etc.


bls20AWS:
Please create an example-pattern.json file.  Use the model template located at https://github.com/aws-samples/serverless-patterns/tree/main/_pattern-model to set up a README, template and any associated code.

SeoyeonPark:
I added example-pattern.json file. I'm sorry I didn't added it before.
Thank you!

PR #931, opened 2022-11-29
Lambda SnapStart pattern

msailes:
*Issue #, if available:*

*Description of changes:*
API GW -> Lambda -> DynamoDB with SnapStart


hustshawn:
Hi @msailes , could you please also add some instruction how to call the APIs? eg. some curl commands could be helpful. Thank you so much.

PR #944, opened 2022-12-15
New serverless pattern - API Gateway REST API to EventBridge with event payload composition and SQS as target

ymwjbxxq:
Description of changes:
API Gateway REST API to EventBridge with event payload composition and SQS as target

The APIGW payload is sent to EventBridge as a custom event payload. In the integration request, a custom payload is created and passed to EventBridge to target an SQS.


ymwjbxxq:
I think the schema validation is failing because the Language is empty
But it is just a SAM template

Issue #950, opened 2022-12-17
Standardise naming of .NET samples


jeastham1993:
On the serverlessland site, .NET examples are found under '.NET', .Net' and 'C#'. Standardise all to use .NET.

PR #952, opened 2022-12-17
Feature/pipes cdk

jeastham1993:
*Issue #, if available:*
#950

*Description of changes:*
Added CDK .NET examples for event bridhge pipes


boyney123:
Thanks @jeastham1993 looking good to me, just a small one, it's easier for us to review if the patterns are split into separate PR, that way we can review faster and also merge faster into ServerlessLand, apart from that all good! Thanks will get live.

PR #958, opened 2022-12-22
New Serverless Pattern: apigw-lambda-json-xml-vtl-transformation

ganvamsi:
*Issue #, if available:*

*Description of changes:*
Added a lambda function and APIGateway mapping template config to convert json response from Lambda to custom XML format.


boyney123:
Thanks @ganvamsi for the contribution. Sorry it took so long to review, we had xmas breaks etc.

I had to make minor changes to the VTL as the XML was not valid coming out, but now works.

ganvamsi:
Thanks @boyney123 for the help . Appreciate the fix.

PR #962, opened 2022-12-27
Rust: upgrade all patterns to the latest version of the aws-sdk for Rust

ymwjbxxq:
*Description of changes:*

- I have updated all the patterns to the latest version of aws-sdk for Rust because, during the year 2022, we got some breaking changes on how the handle is called.
- I also added the example-pattern.json file to align it with the repo rules.


julianwood:
Thanks for the contributing to update a whole bunch!

julianwood:
Thinking of the future, is there a way with Rust to avoid having to specify a version, just select $latest or something? Will avoid manually having to update again?

ymwjbxxq:
Hi Julian,

The Rust version is relatively stable and does not influence the Lambda handler.

In this case, the AWS Team made a breaking change, and it was not Rust but the AWS SDK that changed.

For the libs, for example, it is possible to do this:
Instead of:
`lambda_http = "0.7.2"`
We can set:
`lambda_http = "0.7"`

This will automatically move to the next release.

julianwood:
Thanks, understood. What do you think is best to simplify future maintenance.

ymwjbxxq:
In theory, once you clone it, you can use and deploy with the version of the libs you get, and it should work.
I did update all only for a reason:
- to show the correct way to do it and not the old way

Like the other runtimes, someone should update the lib and the code if something changes drastically.

Imagine a node lambda. We have templates with nodejs14.x, and maybe Node14 will not be supported in the future, and someone should update all. But there is no automatic way. Even moving from node16.x to node18.x require a manual intervention because one requires the SDK v3.

Maybe create a BOT to check this stuff and possibly mark the template as outdated or send an email to the developer that published the pattern to update it kindly

julianwood:
Thanks. Appreciate the info which is useful.

PR #992, opened 2023-01-24
update: AppSync to DDB single table cdk

mtliendo:
Overhauled the repo to make use of both the L2 construct and AppSync JS functions instead of VTL.

note: No `example-pattern.json` file found. Should I create one?


bls20AWS:
Yes, please create an example-pattern.json file.

PR #993, opened 2023-01-24
Updating batch size to 10 for EventBus target

mrbydgn:
*Issue #, if available:*
N/A
*Description of changes:*
Updating batch size to 10 for EventBus target to prevent developers using a non-ideal config


mavi888:
It would be great to have a more clear explanation why to do this update. What is the problem and why 10. 
Thanks

PR #997, opened 2023-01-25
Focusotter-cdk-appsync-sqs

mtliendo:
Update to show L2 construct, pipeline resolver and AppSync JS function


singledigit (approved):
Changes look solid to me.

PR #1007, opened 2023-02-02
Dominik richter feature content filter pattern

dominik-richter:
*Issue #, if available:*

*Description of changes:* Added content filter pattern using pipes with cdk.


content-filter-pattern-kinesis-cdk/src/bin/src.ts
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env node
+import 'source-map-support/register';
+import * as cdk from 'aws-cdk-lib';
+import { ContentFilterStack } from '../lib/content-filter-stack';
```
boyney123, line 4: Think we missing the lib directory @dominik-richter

dominik-richter: Thanks for pointing it out. Changed the gitignore file to allow .js files in lib/lambda

boyney123:
Great pattern! Thanks

PR #1014, opened 2023-02-06
New serverless pattern - dynamodb-streams-lambda-eventbridge-sam-rust

ymwjbxxq:
*Description of changes:*
The AWS SAM template deploys a Lambda function, a DynamoDB table and an Amazon EventBridge bus, and the minimum IAM resources required to run the application.

When items are written in the DynamoDB table, the changes are sent to a stream. This pattern configures a Lambda function to poll this stream. The function is invoked with a payload containing the contents of the table item that changed.


julianwood:
Thanks for the contribution

PR #1017, opened 2023-02-06
appsync-eventbridge-scheduler-sns

mtliendo:
Used the CDK to create an on demand eventBridge scheduler from an AppSync API. The scheduler is configured to publish messages to SNS


mavi888:
@mtliendo thanks for summiting the pattern. But i think this pattern has a very complex way of testing that can be solved with a  couple of simple command lines.

1. A command line to create the subscription to the topic 
`aws sns subscribe \
    --topic-arn arn:aws:sns:us-east-1:XXXX:XXXX \
    --protocol email \
    --notification-endpoint my-email@example.com`
2. A curl command to do the mutation that creates the schedule and then the user can check their email to make sure that the emailed arrive.

Please fix that - you can leave the frontend part, but in the serverless patterns is not the way we try to test things. We want that all the patterns are easily tested mostly from the console. 

Cheers,

mavi888:
@mtliendo If you don't reply to this pr in the next 7 days I will close it. 
Cheers,

mavi888:
Closed due to lack of activity.
@mtliendo if you fix the comments just reopen the PR and we can take it from there.

Cheers,
Marcia

PR #1019, opened 2023-02-07
New serverless pattern - dynamo-pipes-sqs-cdk

twonder:
*Issue #1020*

This pattern demonstrates a low code way to send DynamoDB stream records to an SQS queue using EventBridge Pipes.


singledigit (approved):

singledigit (changes requested):
Tim,

The `example-pattern.json` file is not validating. I believe you are missing the `templateFile` value. Please correct and update.

twonder:
Thanks Eric, I updated the example-pattern.json to include a templateFile which is the C# class that has the stack definition.  Hopefully that works even though it isn't a true template file.

I also tried to run the scripts/validate.js script locally and it seemed like the language C# is supposed to be .NET so I updated that as well.

Are there instructions on running the scripts/validate.js file, or is that not something contributors are supposed to run?  If not, it seems like it would be helpful to put in the README to validate before creating the PR.

PR #1049, opened 2023-02-10
Application dependency upgrade - CRT is now GA, so I've updated the version

msailes:
Also update other dependencies to mitigate CVEs

*Issue #, if available:*

*Description of changes:*


jbesw:
Thanks for the improvement, Mark!

PR #1073, opened 2023-02-16
pushing sfn Lambda cross account pattern

srajaw:
*Issue #, if available:*

*Description of changes: Adding a new pattern to call cross account Lambda function via step function. This pattern will deploy step function and lambda in two other accounts


srajaw:
Not able to identify the exact reason for validation failure. Please help.

PR #1085, opened 2023-02-20
Added pattern for alb to lambda with Pulumi

pierskarsenbarg:
*Issue #, if available:* #1084

*Description of changes:*

Adding a Pulumi pattern to deploy a lambda function behind an ALB

### DIFF
--- a/alb-lambda-pulumi-ts/.gitignore
+++ b/alb-lambda-pulumi-ts/.gitignore
@@ -1,0 +1,2 @@
+/bin/
+/node_modules/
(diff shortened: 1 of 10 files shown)


julianwood:
Thanks for this submission. It's in my review list. As its the first Pulumi one, just taking some extra time to understand how best this should work with Pulumi,

pierskarsenbarg:
@julianwood I didn't want to inundate you with PRs so thought I'd get the first one out the way. 

Happy to work things out on a call or email if that's easier

PR #1119, opened 2023-03-09
feat: eventbridge to sqs pattern (typescript)

ryanseee:
*Issue #, if available:* -

*Description of changes:*

This is the typescript addition to an existing serverless pattern (EventBridge - SQS). It includes an L3 construct which users could also extract and make use as they deem fit.


mavi888:
Thanks for submitting this pattern.
It has been merged and you can see it here: https://serverlessland.com/patterns/eventbridge-sqs-cdk-typescript

Cheers,

PR #1139, opened 2023-03-15
New Pattern: Amazon S3 File Upload using CloudFront and Lambda@Edge

tmmruk:
*Issue #, if available:*

Adding new `cloudfront-lambda-edge-s3-cdk` folder containing new CDK pattern (Typescript) for creating a file upload solution using CloudFrond and Amazon S3, with Lambda@Edge for custom auth. This is to address the 10MB payload limitation with API Gateway S3 proxies, and a simpler solution to the pre-signed URL one.


boyney123:
Thanks @tmmruk

PR #1181, opened 2023-03-31
New serverless pattern - eventbridge-schedule-to-sqs-terraform

f-verg:
*Description of changes:*
This pattern uses EventBridge Scheduler to send messages to an SQS queue every minute.


jbesw:
Your example-pattern.json file has failed validation - please fix this and we will reevaluate the submission.

f-verg:
Should be completed now. I used 'Python' as language even if the code doesn't use any Lambda function with code, as it is only a terraform template.

jbesw:
Thanks for the contribution! This will be live at https://serverlessland.com/patterns/eventbridge-schedule-to-sqs-cdk-terraform shortly.

PR #1269, opened 2023-04-20
Removed hard coded configuration for build EventBridgeClient

ericrlessa:
The first time I deployed the stack an error occurred because the region was different from the hard coded configuration. I just removed the region to build the EventBridgeClient and worked ok.


jbesw:
Thanks for the improvement!

PR #1325, opened 2023-05-02
New serverless pattern - Regional API with Lambda

aliceinaws:
*Issue #, if available:*

*Description of changes:*
New SAM template that creates a Regional Rest API integrated with a Lambda function.
Lambda in python3.9 returns a small message. The inline code is written directly in the template.


jbesw:
Thanks! This will be live at https://serverlessland.com/patterns/apigw-rest-regional-lambda later today.

Issue #1331, opened 2023-05-03
New Pattern Submission : IOT Rule to Lambda


harveyraju:
Subscribe to an IOT Topic and trigger a Lambda using IOT Rule.

When an IOT Topic receives an event, based on the defined IOT rule a lambda will be invoked and the events payload will be sent to the lambda for further processing. This pattern configures the lambda function, IOT rule and add this IOT rule as a trigger for the lambda

PR #1334

PR #1334, opened 2023-05-04
Iotrule lambda trigger pattern

harveyraju:
*Issue #1331 

When an IOT Topic receives an event, based on the defined IOT rule a lambda will be invoked and the events payload will be sent to the lambda for further processing. This pattern configures the lambda function, IOT rule and add this IOT rule as a trigger for the lambda


julianwood:
Thanks for the pattern. I see its in CDK v1. Can this please be migrated to CDK v2 to keep it current

harveyraju:
Hi Julian, I have updated the version from v1 to v2. Please review it.

PR #1387, opened 2023-06-03
Added authors to the example-pattern.json file

harveyraju:
Hi Julian,
I have added the authors details to the example-pattern.json. Please update the same in Serverlessland.
Thanks


jbesw:
Thanks - this change will be live in the next few minutes.

PR #1393, opened 2023-06-07
New Serverless Patterns - Image Compression and Custom Notification upon alarm state change

aroradev-dms:
Added Serverless Pattern for Image Compression and Custom Notification upon alarm state change

1. Image Compression : This pattern leverages the AWS Lambda and S3 event notifications to compress a high quality image from a source bucket to a destination bucket.

2.  Custom Notification upon alarm state change : As we might be aware that notification through CloudWatch Alarm-SNS can not be modified, This pattern focuses on customizing the notifications by leveraging Lambda function. 


singledigit (changes requested):
@aroradev-dms, the best practice is to let CloudFormation name the buckets instead of passing in a Bucket name. Is there a reason we need to on this?

aroradev-dms:
> @aroradev-dms, the best practice is to let CloudFormation name the buckets instead of passing in a Bucket name. Is there a reason we need to on this?

CloudFormation naming the buckets will have same bucket name, thereby asking the users to enter their own source and destination bucket name.

singledigit:
That is actually incorrect. CFN will auto assign a unique name. There is no reason for the user to do this as they are more likely to have a name space collision. Please update.

jbesw:
Hello, this example is more of a solution of a pattern, due to including image compression. I'd recommend implement Eric's fix but submitting to the Serverless Repos instead (https://serverlessland.com/repos).

PR #1434, opened 2023-06-28
Submit a new serverless pattern - API Gateway, Lambda, DynamoDB, Kinesis, Lambda

nguyendinhthi0705:
*Issue #, if available:*

*Description of changes:*

This pattern contains a sample SAM template that deploys a API Gateway will trigger a Lambda Function as backend, then the Lambda will save data into DynamoDB, after that an event will be emit from DynamoDB into Kinesis Stream, Finally, a Lambda will be trigged to proceed new message.


jbesw:
Thanks - this will be live soon at https://serverlessland.com/patterns/apigw-lambda-dynamodb-kinesis-lambda-sam.

PR #1436, opened 2023-06-29
New Serverless Pattern - Adding custom headers using APIGW and Lambda Token Authorizer

interludic2000:
*Description of changes:*

This pattern provides users a way to inject additional information in the form of a HTTP header by using Lambda Authorizer. It can be used in the use-case where downstream legacy systems expect additional data not available from the original API request.


interludic2000:
@singledigit  Are we good to merge this PR?

PR #1452, opened 2023-07-03
New serverless pattern - eventbridge-schedule-to-cloudwatch-sam

professional-ian:
New pattern using SAM for EventBridge Scheduler to custom metrics to CloudWatch a schedule.


julianwood:
Looks like there are multiple patterns here, and I think one you did before.
Could you please split out and have a single pattern per PR which makes it easier to work with.

PR #1457, opened 2023-07-03
Shankey28 feature cdk s3 sqs lambda dynamodb

shankey28:
This pattern provides a simple serverless workflow to process data in a csv file and store in DynamoDB table.

The pattern creates an s3 bucket, an s3 object create event listener a sqs queue that receives messages from the s3 object created event. Lambda polls the sqs queue and process data from the csv file and stores in the DynamoDB table.

Steps to deploy and validate the pattern along with sample data are listed in the Readme file.

### DIFF
--- a/cdk-s3-sqs-lambda-dynamodb/.gitignore
+++ b/cdk-s3-sqs-lambda-dynamodb/.gitignore
@@ -5,4 +5,4 @@ node_modules
 # CDK asset staging directory
 .cdk.staging
 cdk.out
-!data
+!./data/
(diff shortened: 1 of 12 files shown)


julianwood:
looks like you are missing the example-pattern.json file in your pattern.

You can [find the example-pattern template here](https://github.com/aws-samples/serverless-patterns/blob/main/_pattern-model/example-pattern.json).

The file is used on ServerlessLand and is required. Once the file is added we can review the pattern.

PR #1467, opened 2023-07-04
Makendrang feature s3 eventbridge cloudformation

julianwood:
*Issue #, if available:*

*Description of changes:*


julianwood:
I've amended this pattern to use AWS SAM rather than Cloudformation. Same template file but avoids duplication for SAM/Cloudformation templates,

PR #1469, opened 2023-07-05
New serverless pattern - cross account AMI monitoring solution

naveen-amzn:
*Issue #, if available:*

*Description of changes:*
This solution will help customers to create auditing and monitoring solution for cross accounts shared AWS AMI. This will also alert customers when any de-registered AMI is still getting used in any AWS Service. This pattern will help customers to provide insights about AMI usage and reduce any production issue in case any de-registered AMI is referred in customers production account.


jbesw:
Thanks for the submission but this is too complex for a pattern - please upload to a separate repo and submit as a Serverless Land repo at https://serverlessland.com/repos.

PR #1533, opened 2023-07-27
New Serverless Pattern - S3-Lambda-Rekognition-DDB

Chrispyontheoutside:
*Issue #, if available:*

*Description of changes:*
This pattern will give guidance for event driven architectures on classification of images using Rekognition and updating DynamoDB.


julianwood:
Also would be helpful to Output/retrieve  the S3 bucketname from the CDK stack/CLoudformation

Chrispyontheoutside:
Thanks so much - I've updated to include example-pattern.json & to output the bucket name / update README respectively.

PR #1567, opened 2023-08-11
New serverless pattern - apigw-dynamodb-kinesis-lambda

konadevs:
*Issue #, if available:*

*Description of changes:*
REST API that executes a PutItem API call to Amazon DynamoDB table. Amazon DynamoDB table has Amazon Kinesis Data stream to capture item-level changes in your table. The stream is then consumed by a Lambda function.


singledigit:
@konadevs I am a little confused on the need for the Kinesis Data Stream. Why not invoke the Lambda function from the DynamoDB stream?

konadevs:
@singledigit Thank you for looking into this!
Yes, we can directly invoke Lambda from DynamoDB stream.

I worked on a customer use case, where their requirement was to have multiple consumers. With DynamoDB stream we can have only two, however, as they wanted to have multiple consumers I thought of using Kinesis Stream in between.

konadevs:
Hello @singledigit 

Could you please review and take this PR forward?

Thanks

PR #1578, opened 2023-08-19
New serverless pattern - Lambda with container image 

aliceinaws:
*Issue #1579, if available:*

*Description of changes:*
New SAM template that creates a Lambda function running with a container image.
The container image needs to be entered in the parameter of the template from an ECR image already created in the account.


jbesw:
Thanks! This will be live at https://serveressland.com/patterns/lambda-ecr shortly.

PR #1601, opened 2023-08-25
New serverless pattern - Kinesis Data Streams integration with Lambda using the AWS CDK in Java

lefkarag:
This pattern creates an Amazon Kinesis Data Stream that integrates with AWS Lambda using the AWS CDK in Java.


jbesw:
This pattern is missing the example-pattern.json metadata file - please add this to the repo so we can review and test the content.

lefkarag:
The file was already there but I renamed it to `example-pattern.json`.

PR #1641, opened 2023-09-07
Lambda Elastic IP without NAT Gateway

MajdSoufan:
*Issue #, if available:*

*Description of changes:*

This pattern allows you to assign your lambda function a static public IP address that you can use to interact with APIs that require whitelisted IPs without the need to provision a NAT Gateway. Therefore, this pattern will save almost $33/month in NAT Gateway costs.

The following resources will be provisioned:

A Lambda function to test the pattern
An Elastic IP to associate with the Lambda function
A public subnet in the default VPC with us-east-1e as an AZ
A custom resource with Lambda function to associate the Elastic IP with the test lambda's ENI
Since AWS manages the provisioning of any Lambda ENI, we cannot access that ENI in CDK code. Therefore, to automate the process, we have to associate the Elastic IP with the ENI in a custom resource after the deployment occurs.

Disclaimer: This pattern is best suited for non-production environments since it is not multi-AZ nor highly scalable.


MajdSoufan:
@julianwood thank you for taking the time to review my PR. Please, let me know if any changes are required to merge this pattern

PR #1659, opened 2023-09-13
New serverless pattern for multi-country Cognito-SNS use case

rtewani:
*Issue #, if available:*

*Description of changes:*
This pattern will provide guidance on how to use Amazon Cognito’s Lambda triggers to dynamically select origination IDs based on user's country code and send SMS using that specific origination ID.

This pattern uses SAM to build and deploy the architecture to AWS.


rtewani:
> @rtewani looks like you are missing the example-pattern.json file in your pattern.
> 
> You can [find the example-pattern template here](https://github.com/aws-samples/serverless-patterns/blob/main/_pattern-model/example-pattern.json).
> 
> The file is used on ServerlessLand and is required. Once the file is added we can review the pattern.

Added pattern.json file.

PR #1664, opened 2023-09-15
New serverless pattern - Cloudfront to Lambda URL using CDK in TS

samudurand:
New pattern describing how to connect Cloudfront to Lambda via Lambda URL, but using CDK in Typescript, instead of the existing pattern that uses SAM.


jbesw:
Thanks - this will be live at https://serverlessland.com/patterns/cloudfront-lambda-urls-cdk shortly.

PR #1666, opened 2023-09-15
New serverless pattern - apigw-lambda-pinpoint

smalhotraa:
*Issue #, if available:*

*Description of changes:*

This pattern validates the destination number sent in request and allows the users to send SMS through Pinpoint using API Gateway and Lambda. 


jbesw:
Thanks - this will be live at https://serverlessland.com/patterns/apigw-lambda-pinpoint shortly.

smalhotraa:
@jbesw @boyney123 

Hi,

The page doesn't load up the content, it is just blank. Also, when I inspect the same, I see some errors like the following in screenshot. 

Please do let me know if anything has to modified from my end to make it work.

Thanks

Screenshot 2023-09-20 at 2 53 56 PM

PR #1668, opened 2023-09-16
New serverless pattern - Kinesis event source for Lambda using Python CDK

sushir-vr:
*Issue #1669, if available:*

*Description of changes:*
This pattern creates a Kinesis Data Stream that is added as an event source to AWS Lambda function.


jbesw:
Thanks - this will be live at serverlessland.com/patterns/kinesis-lambda-cdk-python shortly.

PR #1676, opened 2023-09-21
apigw-lambda-sns-sms pattern

smalhotraa:
*Issue #, if available:*

*Description of changes:*

This pattern shows an example how a phone number can be validated and how a SMS message can be send using SMS Publish APIs via API Gateway and Lambda deployed using SAM.  


jbesw:
Thanks - this will be live at https://serverlessland.com/patterns/apigw-lambda-sns-sms shortly.

PR #1693, opened 2023-09-28
New Pattern - Kinesis Data Streams to Kinesis Data Firehose to Amazon OpenSeach

biswanathmukherjee:
*Description of changes:*

This SAM template creates a Kinesis Data Stream, Kinesis Data Firehose and Amazon OpenSearch Serverless collection. Kinesis Data Firehose is configured with Kinesis Data Stream as source and Amazon OpenSearch as destination. When a message is pushed to Kinesis Data Stream, the same is replicated to OpenSearch using Kinesis Data Firehose. If any messages that can not be processed by Kinesis Data Firehose the same is sent to S3 bucket.


jbesw:
Thanks - this will be live at https://serverlessland.com/patterns/kinesis-data-firehose-firehose-opensearch-sam shortly.

PR #1706, opened 2023-10-02
New serverless pattern - CloudFront to S3 with OAC

lctnstn:
This pattern creates a CloudFront distribution with an S3 origin secured with origin access control (OAC).


singledigit (changes requested):
Please update the template to dynamically name the Bucket and reference it. Since buckets are global and uniques, this could cause a clash and fail.

lctnstn:
@singledigit -- changed the bucket name to be account id + deployment name to ensure uniqueness. All good?

lctnstn (commented):
All set.

PR #1716, opened 2023-10-06
New serverless pattern - Amazon Aurora Serverless V2 Primary to Secondary Global Database

nihilg:
*Issue #, if available:*

*Description of changes:*

The pattern creates a serverless global database cluster enabling data replication from the primary to secondary cluster. The regional primary cluster contains a serverless db instance supporting both writes and reads. The regional secondary cluster contains a serverless db instance supporting only reads. In the unlikely event of a regional degradation or outage, the secondary region can be promoted to read and write capabilities in less than 1 minute. A fargate based application is also provisioned to test the replication. Also the pattern adopts the multiple stack capability of CDK to provision the resources across the primary and secondary regions. You can deploy each stack individually or deploy all the stacks using --all option.

### DIFF
--- a/aurora-serverless-global-db-cdk/.gitignore
+++ b/aurora-serverless-global-db-cdk/.gitignore
@@ -1,0 +1,10 @@
+*.js
+!jest.config.js
+*.d.ts
+node_modules
+.DS_Store
+package-lock.json
+
+# CDK asset staging directory
+.cdk.staging
+cdk.out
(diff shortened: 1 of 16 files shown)


mavi888:
Thanks for submitting your pattern.
Now this pattern is public in Serverless land

https://serverlessland.com/patterns/aurora-serverless-global-db-cdk

You can share your pattern using this link :)

PR #1765, opened 2023-10-18
New Serverless Pattern - Priority Queue using SQS

biswanathmukherjee:
*Description of changes:*

This sample project showcases Priority Queue Pattern using AWS Serverless services. There are three priorities - High, Medium and Low. The project demonstrates the behavior using both standard and FIFO SQS Queues. Lambda is used as consumer of the messages from SQS queues. The objective is to first process the messages from High priority queue and only if there is no messages in High priority queue, process from Medium and subsequently from Low priority queue if no more messages left on High and Medium priority queues. As it is observed in the results, FIFO queue performs much better in maintaining the priority however Standard queue might not be able to maintain the priority in some instances as the ordering of the messages is on best effort basis.


julianwood:
Hi, thanks for the submission. This pattern seems overly complex.
I suggest adding this as a repo on aws-samples which we can link to via the repo submissions.
Also, this patterns could have issues if the queues scale. If it takes longer than 15 minutes to drain all queues, for example, this will fail.
I'm not sure this is a good architectural pattern, DM me if you'd like to discuss further.

julianwood:
Closing to move to repo

PR #1788, opened 2023-10-23
New serverless pattern - Marketplace metering records entry and listing using APIGateway 

saikotag:
*Issue #, if available:*

*Description of changes: AWS Marketplace Metering information that needs to be reported by sellers should be stored in DynamoDB table called AWSMarketplaceMeteringRecords. This table gets created in the Seller’s AWS account as part of AWS Markerplace [SaaS quickstart integration.](https://catalog.workshops.aws/mpseller/en-US/saas/integration-with-quickstart) 
Issue: Currently, sellers manually insert a record into AWSMarketplaceMeteringRecords table either using DynamoDB API or someother client. 
Solution: This serverless patter will create API Gateway Rest endpoint which can be used to insert these metering records into DynamoDB table. These endpoints can also be used to automate the record sending.


jbesw:

PR #1790, opened 2023-10-24
New serverless pattern - cognito-whatsapp-otp

ReethiJ:
*Issue #, if available:* [1791](https://github.com/aws-samples/serverless-patterns/issues/1791)

*Description of changes:*

This pattern provides guidance on how to use an Amazon Cognito Userpool's custom SMS sender Lambda trigger to send OTP messages on WhatsApp. It is a TypeScript CDK template that creates a Amazon Cognito userpool, AWS KMS key and AWS Lambda function which is the custom sms sender Lambda. Lambda is in Python 3.11.


jbesw:

PR #1827, opened 2023-11-01
Serverless RAG #1826

giusedroid:
*Issue #, if available:* #1826

With this pattern we want to showcase how to implement a serverless Retrieval Augmented Generation (RAG) architecture.  
Customers asked for a way to quickly test RAG capabilities on a small number of documents without managing infrastructure for contextual knowledge and non-parametric memory.  
In this pattern, we run a RAG workflow in a single Lambda function, so that customers only pay for the infrastructure they use, when they use it.  
This pattern makes use of Bedrock to calculate embeddings with Amazon Titan Embedding and Claude as prediction LLM.  
We also provide a local pipeline to ingest your PDFs and upload them to S3.

Authors: giusedroid, shafkevi


jbesw:

PR #1832, opened 2023-11-02
Streaming Serverless RAG with Amazon Bedrock

shafkevi:
Issue #, if available: https://github.com/aws-samples/serverless-patterns/issues/1831

With this pattern I want to showcase how to implement a streaming serverless Retrieval Augmented Generation (RAG) architecture.
Customers asked for a way to quickly test RAG capabilities on a small number of documents without managing infrastructure for contextual knowledge and non-parametric memory.
In this pattern, I run a RAG workflow in a single Lambda function, so that customers only pay for the infrastructure they use, when they use it.
The responses are streamed using Lambda URL function streaming for a quicker time to first byte and a better user experience.
This pattern makes use of Bedrock to calculate embeddings with Amazon Titan Embedding and any Amazon Bedrock chat model as prediction LLM.
I also provide a local pipeline to ingest your PDFs and upload them to S3.
There is also an example of how to consume streamed responses using a very basic React application.

Authors: shafkevi


jbesw:

shafkevi:
Resubmitted as a repo: https://github.com/aws-samples/serverless-patterns/issues/1836

PR #1849, opened 2023-11-07
New serverless pattern - account-tagging-at-scale

pmvrak:
*Description of changes:*

This pattern helps to tag accounts at scale. This solution is complete serverless in nature and is built on `python` and `Terraform`. This solution takes user input as tag key/value pairs. The solution builds an AWS Lambda Funtion that takes the tag input data (key/value pair) and uses AWS Organization API's to tag each and every account mentioned in the user input file. See `Architecture` section for the architecture flow and `Deployment Instructions` for detailed information.


jbesw:
Hello - please submit this as a repo (https://serverlessland.com/repos), not a pattern. Patterns are intended as 2-3 services with minimal code that are the basis for building new applications.

PR #1852, opened 2023-11-07
feat: Disable EventBridge rules with EventBridge Scheduler

embano1:
*Issue #, if available:* Closes: #1842

*Description of changes:* New pattern


jbesw:
Hello - please submit this as a repo (https://serverlessland.com/repos), not a pattern. Patterns are intended as 2-3 services with minimal code that are the basis for building new applications.

boyney123:
Just reviewed this, think its a simple schedule that edits rules, I'm happy to own reopen this and add it to patterns, we have other schedules similar in size, so think pattern will be OK here.

PR #1869, opened 2023-11-11
add dynamodb-stream-opensearch-serverless pattern CDK

fanhongy:
*Issue #, if available:*

*Description of changes:*

This pattern allows you to stream new created item on a specific DynamoDB talbe to a Opensearch serverless collection for searching capability for DyanmoDB table.


boyney123:
Missing example-json file and also you have 512 files changed in the pull request, that doesn't sound right? Tried to see which files changed on GitHub, but it crashes the browser..... can you verify, and add the missing example file?

fanhongy:
Thanks for you response. I will add the exaple-json file, and update the pull request. For 512 files added, this is due to a Lambda layer requires dependencies which also part of the pattern.

jbesw:
This PR is too large/complicated for a pattern at over 500 files. I recommend you submit as a repo instead (https://serverlessland.com/repos).

PR #1897, opened 2023-11-21
New serverless pattern - Ingesting CloudWatch Logs into OpenSearch Serverless

skyjasonaws:
*Issue #, if available:* #1896

*Description of changes:*
This solution creates an automated pipeline to ingest CloudWatch Logs into OpenSearch Serverless for analysis. A series of AWS services are orchestrated to form a flexible, reliable data flow from source logs to destination search and analytics engine.  The key component is a Lambda function that transforms the data between managed service formats.


jbesw:
This submission would be better fit for the Repos Collection (see https://serverlessland.com/repos/). Patterns is intended for skeleton IaC apps of 2-4 AWS services with minimal business logic or custom logic. Since this has 200+ lines of code and more services, it fits the repos model instead. If you can resubmit there, we will get it merged as soon as possible.

PR #1898, opened 2023-11-22
Added API Gateway REST API to Lambda to SNS in nodejs 

higoyal01:
*Issue #, if available:*

*Description of changes:*
Added API Gateway REST API to Lambda to SNS in nodejs 


higoyal01:
Hi Team, do we have an update on this?

boyney123:
Hi @higoyal01 

Sorry we just got back from re:Invent 2023, and getting back to speed, thanks for the patience.

singledigit:
Hi @higoyal01 thank you for submitting the pattern. However, there are several things that go against best practices:
1. The pattern is using a full OpenApi spec without reason. This could be handled by a simple API event on the Lambda function.
2. The pattern uses a custom policy when a sam policy will do
3. Additionally the custom policy uses '*' in the resource for the SNS topic. This makes it too broad of scope.

Finally, this isn't really a good fit for patterns as it is an architectural example of API gateway to Lambda, which we already have plenty of. It's the code that connects to the SNS topic and differentiates this submission. If you would like to resubmit as a snippet on the code please go here: https://serverlessland.com/snippets

Closing.

PR #1900, opened 2023-11-22
API Gateway to Lambda to Comprehend using SAM and Boto3

abilashkumar:
*Description of changes:*

This pattern contains the application code in Python and supporting files for a serverless application that you can deploy using the SAM CLI. It creates a REST API (API Gateway) with a Lambda function as a back end integration that uses Python (Boto3 SDK) which calls the Comprehend service to perform real-time detection of the dominant language of the input text.


singledigit:
Hi @abilashkumar, this is not a good candidate for a pattern. A pattern is an architectural example of connecting services together. This demonstrates API Gateway to Lambda using python, which we have many examples of. Where this gets interesting is the code to interact with Comprehend. I would resubmit this as a snippet in the snippets library. https://serverlessland.com/snippets. Closing.

PR #1901, opened 2023-11-22
API GW to Lambda to Comprehend using SAM and Nodejs for language detection

abilashkumar:
*Description of changes:*

This pattern contains the application code in Javascript and supporting files for a serverless application that you can deploy using the SAM CLI. It creates a REST API (API Gateway) with a Lambda function as a back end integration that uses Javacript (v3 SDK) which calls the Comprehend service to perform real-time detection of the dominant language of the input text.

### DIFF
--- a/apigw-lambda-comprehend-language-sam-nodejs/template.yaml
+++ b/apigw-lambda-comprehend-language-sam-nodejs/template.yaml
@@ -1,0 +1,41 @@
+AWSTemplateFormatVersion: '2010-09-09'
+Transform: AWS::Serverless-2016-10-31
+Description: A REST API is created with lambda back end integration that detects the dominant language in the provided input text using Amazon Comprehend.
+Resources:
+  DetectLanguageApi:
+    Type: AWS::Serverless::Api
+    Properties:
+      StageName: dev
+      OpenApiVersion: '2.0'
+
+  DetectLanguageLambdaFunction:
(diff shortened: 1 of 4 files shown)


apigw-lambda-comprehend-language-sam-nodejs/template.yaml
```diff
@@ -0,0 +1,41 @@
...
+        - arm64
+      Timeout: 10
+      Policies:
+        - Version: '2012-10-17'
+          Statement:
+          - Sid: LambdaDetectDominantLanguage
+            Effect: Allow
+            Action:
+            - comprehend:DetectDominantLanguage
+            Resource: '*'
```
boyney123, line 28: Hey @abilashkumar same as the other pattern I reviewed, is the * resource required here? Can we lock this down at all?

abilashkumar: @boyney123 This is required as the permission here is not tied to a resource, but in general to perform language detection on free text provided as an input. The "*" is required (it does not involve any customer resource or their saved configurations). A similar pattern published [earlier](https://github.com/aws-samples/serverless-patterns/blob/main/lambda-translator/template.yaml#L20).

jbesw:
Due to the large amount of custom code, this should be submitted as a repo instead of a pattern: https://serverlessland.com/repos/.

PR #1905, opened 2023-11-22
update the config.json file to get the right template file

lucamezzalira:
When the template was published on serverlessland.com I realized I pointed to the wrong file in the ```config.json``` so with this PR we should display the CDK template instead of the lambda handler

*Issue #, if available:*

*Description of changes:*


lambda-node-esm-cdk/lambda-node-esm-cdk.json
```diff
@@ -16,7 +16,7 @@
         "repoURL": "https://github.com/aws-samples/serverless-patterns/tree/main/lambda-node-esm-cdk",
         "templateURL": "serverless-patterns/lambda-node-esm-cdk",
         "projectFolder": "lambda-node-esm-cdk",
-        "templateFile": "src/hello-world.ts"
+        "templateFile": "lib/node-esm-stack.ts"
```
boyney123, line 19:
```suggestion
        "templateFile": "bin/node-esm-stack.ts"
```

Thanks @lucamezzalira 

Is this meant to be the bin dir? https://github.com/aws-samples/serverless-patterns/tree/main/lambda-node-esm-cdk/bin

lucamezzalira: not really, in the bin I got the stack initialization but the configuration lives inside lib.
This is the default configuration that you have when you run ```cdk init app --language typescript```
In other examples I use to configure in bin the initialization and env params and in lib folder all my stacks

PR #1910, opened 2023-11-23
Minor: Consistent `node:` prefix in imports for `lambda-node-esm-cdk`

lmammino:
*Description of changes:*

Uses the `node:` prefix also for the `path` import

PS: thanks for the awesome example, @lucamezzalira! ❤️


lucamezzalira:
I always forgot about path package 😂 thanks for catching it

PR #1927, opened 2023-12-01
New serverless pattern - apigw lambda comprehend cdk python

awsarippa:
*Issue #, if available:*

*Description of changes:*
The purpose of this PR is to showcase a serverless pattern of CDK deployment for Sentiment Analysis using Lambda, Comprehend, and API Gateway.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1939, opened 2023-12-06
Apigw lambda comprehend cdk python

julianwood:
*Issue #, if available:*

*Description of changes:*


apigw-lambda-comprehend-cdk-python/src/lambda_function.py
```diff
@@ -0,0 +1,37 @@
...
+        LOG.error("Sentinment detection failed!!")
+        raise
+    else:
+        analyzed_response = json.dumps(
+            {
+                "Sentiment": response["Sentiment"],
+                "Confidence Score": response["SentimentScore"],
+            }
+        )
+        return analyzed_response
```
embano1, line 37: Question: is there a reason why this code is not in the `try` block?

embano1: Hm, not sure I follow.

Can you explain why the equivalent w/out elese doesn't this work here? Also, any other exception thrown in your code is not handled (e.g., dict key errors). Is that intentional? If not, consider making exception handling more general (catch any exception, log, and just raise)

```python
    try:
        LOG.info(f"Event is {event}")
        LOG.info(event["body"])
        user_input = json.loads(event["body"])["input"]
        response = comprehend_obj.detect_sentiment(Text=user_input, LanguageCode="en")
        LOG.info(response)
        LOG.info(f"Sentiment detect is {response['Sentiment']}!!")
        analyzed_response = json.dumps(
            {
                "Sentiment": response["Sentiment"],
                "Confidence Score": response["SentimentScore"],
            }
        )
        return analyzed_response
    except ClientError:
        LOG.error("Sentinment detection failed!!")
        raise
```

awsarippa: @embano1 - Your assumption is right. It was not intentional. The idea was to execute return the response through else block. 

I've modified the code to catch any exception in the original PR# https://github.com/aws-samples/serverless-patterns/pull/1927

awsarippa (commented):
Modified try-catch exception block for catching general exceptions.

PR #1961, opened 2023-12-15
New serverless pattern pinpoint-apigateway-amplify-lambda

jainniha:
*Issue #, if available:*

*Description of changes:*

This pattern is sending web push notifications by using amplify, pinpoint, api gateway and lambda. 


jbesw:
Hi - this should be submitted to the Repos collection instead at https://serverlessland.com/repos. Patterns are intended to be small, low-code IaC integrations only.

PR #1967, opened 2023-12-18
Fix: modify Serverlessland URL path and update diagram label to WebSocket API

windrichie:
*Issue #, if available:*

*Description of changes:*
- modified API Gateway label to WebSocket and renamed the JSON file

### DIFF
--- a/apigw-websocket-api-bedrock-streaming/apigw-websocket-api-bedrock-streaming.json
+++ b/apigw-websocket-api-bedrock-streaming/apigw-websocket-api-bedrock-streaming.json
@@ -59,7 +59,7 @@
             "x": 20,
             "y": 40,
             "service": "apigw",
-            "label": "API Gateway REST API"
+            "label": "API Gateway WebSocket API"
         },
         "icon2": {
             "x": 50,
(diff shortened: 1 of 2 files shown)


cornelcroi:
example-pattern.json is missing. Can you add it please?
 Also, I don't see any modification regarding the URL path.

windrichie:
@cornelcroi the example-pattern.json file is already in the current repo: https://github.com/aws-samples/serverless-patterns/blob/main/apigw-websocket-api-bedrock-streaming/example-pattern.json. That's why it is not in this PR.

For the URL path, I renamed the apigw_lambda_bedrock.json to apigw-websocket-api-bedrock-streaming.json, as I thought that was where the URL path is taken from. In any case, I think `apigw-websocket-api-bedrock-streaming` would be a suitable path.

cornelcroi:
Actually `example-pattern.json` needs to be in the PR as we have a github action that looks for it. Can you just make a mior change so that the file is part of the PR?

windrichie:
got it, made a minor change to example-pattern.json @cornelcroi

cornelcroi (approved):

PR #1979, opened 2023-12-22
dyanmodb-lambda-dynamodb (multi-region data replication)

saiaunghlyanhtet:
## Pattern Overview
The terraform template deploys a Lambda function, two DynamoDB in different regions, and the minimum IAM resources required to run the application.

When items are written or updated or deleted in the table in first region (e.g. us-east-1), the changes are sent to a stream. This pattern configures a Lambda function to poll this stream. When the function is invoked, it retrieves data from the payload containing the contents of the table item that has been written or updated or deleted in the table in the first region, then it makes changes in the table of the second region (e.g. us-west-2) based on the event type.


bfreiberg:
Maybe I'm missing something but what is the difference to the build-in feature of Global Tables? https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/GlobalTables.html

saiaunghlyanhtet:
> Maybe I'm missing something but what is the difference to the build-in feature of Global Tables? https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/GlobalTables.html

The pattern is mostly the same as the built-in feature of Global Tables, but in the case of Global Tables, they offer eventual consistency, not strong consistency for writes as far as I know. But that problem can be resolved by using dynamodb auto scaling. Overall, the only difference is that the pattern helps us achieve strong consistency for writes without the use of auto scaling.

parikhudit:
@saiaunghlyanhtet could you kindly check this [prompt](https://github.com/aws-samples/serverless-patterns/pull/1979#issuecomment-1878690729) about example-pattern.json and make suggested updates?

bfreiberg:
> The pattern is mostly the same as the built-in feature of Global Tables, but in the case of Global Tables, they offer eventual consistency, not strong consistency for writes as far as I know. But that problem can be resolved by using dynamodb auto scaling. Overall, the only difference is that the pattern helps us achieve strong consistency for writes without the use of auto scaling.

Could you please explain how this pattern achieves strong consistency? Using a Lambda function to replicate the data across regions naturally introduces some latency. Meaning that updates are propagated with some (even if minor) delay.

parikhudit:
> > The pattern is mostly the same as the built-in feature of Global Tables, but in the case of Global Tables, they offer eventual consistency, not strong consistency for writes as far as I know. But that problem can be resolved by using dynamodb auto scaling. Overall, the only difference is that the pattern helps us achieve strong consistency for writes without the use of auto scaling.
> 
> Could you please explain how this pattern achieves strong consistency? Using a Lambda function to replicate the data across regions naturally introduces some latency. Meaning that updates are propagated with some (even if minor) delay.

@saiaunghlyanhtet closing the PR for now. Feel free to reopen & respond.

PR #1997, opened 2023-12-30
feat(apigw-http-api-eventbridge-lambda-sls): Modernize the example

pgrzesik:
*Issue #, if available:* N/A

*Description of changes:*

When browsing examples on Serverless Patterns website, I've noticed that some of them could be slightly simplified or improved. This PR includes changes for `apigw-http-api-eventbridge-lambda-sls` pattern such as:

- Update of runtime to `nodejs20.x`. Current `nodejs14.x`-based example is not deployable
- Remove unnecessary override of `provider.region` as the provided override configuration is the same as the default
- Added `deploymentMethod` to make the deployment faster by using direct CloudFormation `updateStack` instead of going through changesets which are slower
- Removed unnecessary `memorySize` on function level as it's already set on `provider`
- Bump versions of dependencies

Please let me know what do you think 🙇 

### DIFF
--- a/apigw-http-api-eventbridge-lambda-sls/package.json
+++ b/apigw-http-api-eventbridge-lambda-sls/package.json
@@ -4,8 +4,8 @@
   "license": "MIT-0",
   "type": "module",
   "devDependencies": {
-    "@types/aws-lambda": "^8.10.93",
-    "serverless-plugin-typescript": "^2.1.1",
-    "typescript": "^4.6.3"
+    "@types/aws-lambda": "^8.10.130",
+    "serverless-plugin-typescript": "^2.1.5",
+    "typescript": "^4.9.5"
   }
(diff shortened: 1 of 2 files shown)


cornelcroi (approved):

bls20AWS:
@pgrzesik i do not see a pattern named apigw-http-api-eventbridge-lambda-sls.

pgrzesik:
Hey @bls20AWS 👋 Sorry for missing your comment here and coming back to it just now, I see you reverted this commit, could you please share what is the reason for that? I've noticed now that there's a requirement for `example-pattern.json` in new patterns.

Am I understanding correctly that the pattern this PR is for is kinda "broken" due to missing `example-pattern.json` and that should be addressed first? Thanks in advance 🙇

PR #2013, opened 2024-01-08
New pattern - APIGW Lambda with wildcard resource-based policy using CDK

sabornibhattacharya:
*Issue #, if available:*

*Description of changes:*

Create a REST API with proxy integration to a Lambda function to keep the size of resource-based policy within the allowed hard limit. Currently, in an API Gateway-Lambda setup whenever an user adds an integration, CDK adds a new policy to the Lambda function's resource-based policy. It might result in exceeding the policy size limit for Lambda function which is 20KB. This sample project demonstrates how to use CDK to create a customized integration that would keep the policy size within limit by using wildcards in the resource-based policy.


cornelcroi (approved):

cornelcroi:
> I'm not seeing the updated patterns json file @cornelcroi

Oups. Done.

PR #2028, opened 2024-01-10
New serverless pattern - AWS Step Functions wait for callback functionality using SAM

biswanathmukherjee:
*Issue #, if available:*

*Description of changes:*

The SAM template deploys a Step Functions Standard workflow that takes a JSON input (for example, payload to charge customer). The state machine sends the message to Amazon SQS Queue and waits for the callback. An AWS Lambda function is deployed to process the message from the SQS Queue and calls the state machine with either task success or failure. Depending on the callback type (success or failure) the state machine publishes messages to Amazon SNS Topic for notification. The SAM template contains the required resouces with IAM permission to run the application with logging enabled.


parikhudit:
Thanks for the PR @biswanathmukherjee  
Unfortunately, we have similar pattern here: https://serverlessland.com/workflows/wait-for-callback.

biswanathmukherjee:
Thanks for the link. As it seems similar pattern already exists, I am closing this PR.

PR #2074, opened 2024-01-28
New Serverless Patter - Model customisation in Amazon Bedrock using AWS Step Function

biswanathmukherjee:
*Issue #, if available:*

*Description of changes:*
The SAM template deploys a Step Functions workflow that creates custom model by fine tuning Amazon Bedrock Cohere Command Light model and then creates a provisioned throughput with that custom model. At the end we test both the base model and the fine-tuned model to compare results.


pputhran:
Please submit this pattern to the [Workflow pattern](https://serverlessland.com/workflows).

Issue #2080, opened 2024-02-02
README has invalid JSON has input parameter


madhavvishnubhatta:
In this pattern - eventbridge-pipes-sqs-to-stepfunctions-cdk-python, the [README file](https://github.com/aws-samples/serverless-patterns/tree/main/eventbridge-pipes-sqs-to-stepfunctions-cdk-python) has a command to send a message to the SQS queue to test the pattern. But the command is incorrect as the message body is not a proper json.

```
 aws sqs send-message \
 --queue-url=SQS_URL \
 --message-body '{"orderId":"125a2e1e-d420-482e-8008-5a606f4b2076, "customerId": "a48516db-66aa-4dbc-bb66-a7f058c5ec24", "type": "NEW"}'
```

The order id value's end quote is missing. So SQS things this is a string message body. So the step function does not read the order id and customer id correctly. Needs to be changed to add the missing double quote as below.

```
 aws sqs send-message \
 --queue-url=SQS_URL \
 --message-body '{"orderId":"125a2e1e-d420-482e-8008-5a606f4b2076", "customerId": "a48516db-66aa-4dbc-bb66-a7f058c5ec24", "type": "NEW"}'
```

PR #2081, opened 2024-02-02
Modified the aws sqs send-message command to properly close the value…

madhavvishnubhatta:
… in the json paramter

*Issue #, if available:*
#2080 

*Description of changes:*
The `aws sqs send-message` command to test the pattern has an invalid json as value for the `message-body` parameter. This fixes that by adding a double quote


madhavvishnubhatta:
> @undefined looks like you are missing the example-pattern.json file in your pattern.
> 
> You can [find the example-pattern template here](https://github.com/aws-samples/serverless-patterns/blob/main/_pattern-model/example-pattern.json).
> 
> The file is used on ServerlessLand and is required. Once the file is added we can review the pattern.

@ellisms Do I need to do anything about the above message? Seems like the message is more relevant for new patterns, not for fixes like these.

PR #2084, opened 2024-02-03
EC2 to Amazon S3 Express One Zone Directory Bucket

biswanathmukherjee:
*Issue #, if available:*

*Description of changes:*
SAM Template that creates an EC2 Instance, an Amazon S3 Express One Zone directory bucket and required IAM Role to access the bucket from the instance.


ec2-s3-express-one-zone-sam/.gitignore
parikhudit, line 1: Please remove this file. The base directory of repo already has one.

biswanathmukherjee: Removed.

parikhudit:
Thanks for the contribution, a DA would review it as next step. 

For later, if you could demonstrate or add reference of performance comparison using S3 Express One Zone and S3 standard, it would be a good value addition.

mavi888:
Thanks for submitting your pattern.
Now this pattern is public in Serverless land

https://serverlessland.com/patterns/ec2-s3-express-one-zone-sam

You can share your pattern using this link :)

PR #2153, opened 2024-03-05
New Serverless Pattern - Flink with Data Stream Source and S3 Sink CDK

abdulahad95:
*Issue #, if available:*
#2154 

*Description of changes:*
CDK code for a new serverless pattern - Flink with Data Stream Source and S3 Sink


cornelcroi:
Hi,

This would be more suitable for inclusion in the repository collection at https://serverlessland.com/repos. 
A diagram 
You should submit it there and don't forget to include a diagram as well.

abdulahad95:
Hi @cornelcroi can you explain why this might not be suitable for this repo?

This is the same pattern as [this one](https://serverlessland.com/patterns/kinesis-dataanalytics-flink-s3-sink) except that it is written in Typescript instead of Terraform. 

Also, it is a very similar pattern to [this one](https://serverlessland.com/patterns/cdk-kinesis-flink-firehose) except that it has an S3 sink instead of a Firehose sink.

It looks like the pattern is in line with existing patterns, so I'm not sure why it's not suitable for this repo.

PR #2165, opened 2024-03-07
New serverless pattern - awstransfer-lambda-sam

jackdwayne:
This pattern in SAM offers an example to setup an AWS SFTP server using AWS Transfer Family for SFTP with a custom workflow step to an AWS Lambda Function. The template enables you to simply create a server, a custom workflow, set up user accounts, and an Amazon Simple Storage Service (Amazon S3) bucket. You have fine-grained control over user identity, permissions, and keys. You can create users within Transfer for SFTP. You can also use IAM policies to control the level of access granted to each user. When a user uploads a file to the SFTP server it triggers a custom document processing step.


transfer-lambda/example-pattern.json
```diff
@@ -0,0 +1,57 @@
+{
+  "title": "AWS Transfer Family SFTP Server to AWS Lambda",
+  "description": "This pattern in SAM offers an example to setup an AWS SFTP server using AWS Transfer Family for SFTP with a custom workflow step to an AWS Lambda Function.",
```
parikhudit, line 3: Length needs to be less than 150 chars.

transfer-lambda/example-pattern.json
```diff
@@ -0,0 +1,57 @@
...
+  "language": "Python",
+  "level": "200",
+  "framework": "SAM",
+  "introBox": {
+    "headline": "Create a AWS Transfer Family SFTP Server to custom workflow step Lambda function.",
```
parikhudit, line 8: Length needs to be less than 75 chars.

parikhudit:
For a pattern that uses SAM to deploy solution for S3 to Lambda to DynamoDB, I'd suggest naming it as ```s3-lambda-dynamodb-sam```. For this case, I'd suggest renaming the folder name from ```transfer-lambda``` to something like ```awstransfer-lambda-sam```.

jackdwayne:
Suggested changes have been made to folder and naming conventions in the README and example-pattern.json

PR #2173, opened 2024-03-08
fix: web acl association race condition

JamesKyburz:
*Fix race condition when creating a WAF ACL Association resource:*

CloudFormation expects the api gateway stage to be created, to fix this we can Reference our Api Gateway Stage resource using ${logicalId.Stage}

This fixes an issue I got when testing the proposed template

```
AWS WAF couldn?t perform the operation because your resource doesn?t exist.
```


bfreiberg:
Looks good, thanks for the fix

PR #2193, opened 2024-03-22
Updating NodeJS version to supported runtime

tmmruk:
*Issue #, if available:*

*Description of changes:* Bumping NodeJS Lambda runtime from 14.x to 18.x as 14.x is no longer supported and causes cdk deploy to fail.


bfreiberg:
Due to the replication, either the function needs to be deleted some time after the distribution has been deleted or it needs to be disassociated before deletion. What do you think is the better approach?

tmmruk:
I don't think there is any other option than to wait and re-run cdk destroy after some time. "Replicas are typically deleted within a few hours. You cannot manually delete Lambda@Edge function replicas. This helps prevent a situation where a replica is deleted that is still in use, which would result in an error."

bfreiberg:
Fair point and thank you for your contribution

PR #2195, opened 2024-03-23
Add a pattern showing using compiled dependencies for ruby

joyofhex:
*Issue #, if available:*

*Description of changes:* This adds a pattern that can be used when you need to use a gem with native extensions that also require shared libraries to be available at runtime. To do so we build a custom build container containing the needed libraries, then create a lambda layer that has those libraries present. Finally we build the application and create a .zip archive.


jbesw:
Please submit this to the repos directory at serverlessland.com/repos.

PR #2269, opened 2024-04-29
New Serverless Pattern - Amazon Cognito user management using Amazon API Gateway and AWS Lambda (Java)

biswanathmukherjee:
*Issue #, if available:*

*Description of changes:*
This sample project deploys an Amazon Cognitor user pool, an Amazon API Gateway REST API with an AWS Lambda integration. The Lambda function, written in Java, calls the Amazon Cognito API to create and confirm a user. The project also exposes API to return tokens after successful login by the user


jbesw:
While this is useful, it should be submitted as a repo and not a pattern, due to the amount of custom code. See https://serverlessland.com/repos/ for submission info.

PR #2285, opened 2024-05-19
New serverless pattern - s3-lambda-bedrock-js-sam

Avs163:
*Issue #, if available:*

*Description of changes:*

This pattern facilitates LLM RAG use cases by generating vector embeddings of documents in S3. When files are uploaded to S3, Lambda triggers, sending file contents to the Bedrock embedding model, which generates semantic vector embeddings for NLP tasks like semantic search, document clustering, or information retrieval.


s3-lambda-bedrock-js-sam/template.yaml
```diff
@@ -0,0 +1,76 @@
...
+                Action:
+                  - bedrock:InvokeModel
+                Resource: "*"
+
+Outputs:
+  DocumentBucket:
+    Description: S3 bucket for storing documents
+    Value: !Ref DocumentBucket
+  DocumentEmbeddingFunction:
+    Value: !Ref DocumentEmbeddingFunction
```
parikhudit, line 75: This would not return an ARN but Function Name

s3-lambda-bedrock-js-sam/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+When a new object is uploaded to an S3 bucket, S3 detects the putObject event and invokes a Lambda function asynchronously, passing details about the uploaded object. The Lambda function then retrieves the object's content and sends it to the Amazon Bedrock Titan Embedding service, which generates vector embeddings representing the content's meaning and context.
+
+## Testing
+
+To test the system, we should first tail the logs of the Lambda function. Then, we can upload the sample.txt to our designated S3 bucket. Once the document is uploaded, we can observe the logs to verify the Lambda function's execution and the embeddings of the uploaded document. 
+
+1. Tail to the Lambda function log to get the logs
+```
+aws logs tail "/aws/lambda/DocumentEmbeddingFunction" --follow --format json
```
parikhudit, line 50: The function name ```DocumentEmbedding``` in CloudFormation output would not match ```DocumentEmbeddingFunction ```

s3-lambda-bedrock-js-sam/template.yaml
Avs163: Made changes to description.

s3-lambda-bedrock-js-sam/README.md
Avs163: Made the changes. Corrected the name to DocumentEmbedding.

s3-lambda-bedrock-js-sam/template.yaml
parikhudit: Thanks for adding description. However, I meant that this output will not have function ARN, it will return Function Name. To get ARN, you may need to do something like ```!GetAtt DocumentEmbeddingFunction.Arn``` in value. Please test it once.

parikhudit: Thanks for adding description. However, I meant that this output will not have function ARN, it will return Function Name. 

If you need to return ARN, you may need to do something like ```!GetAtt DocumentEmbeddingFunction.Arn``` in value. 
If you need Function Name, keep it as it is but update the description.

Avs163: We are returning the name of the lambda function, and we have added the required description that, `!Ref DocumentEmbeddingFunction` will return Lambda function name, while `!GetAtt DocumentEmbeddingFunction.Arn` will return ARN for the lambda. Please see the latest commit made to the PR.

PR #2287, opened 2024-05-28
Modified Python lambda runtimes from 3.7 to 3.12

keithrozario:
*Issue #, if available:*

*Description of changes:*
Multiple Patterns will no longer work because the AWS Lambda Runtime was set to `python3.7`. In some cases it would be difficult to change as the solution incorporates their own packages/libraries that were built for this version of python (for example by pulling in source from an s3 bucket). 

I went through most examples, and where the solution was simple, changed the python runtime to Python3.12, to give us more runtime before we need to change these again. 

### DIFF
--- a/alb-cognito-lambda/template.yaml
+++ b/alb-cognito-lambda/template.yaml
@@ -64,7 +64,7 @@ Resources:
       FunctionName: "ALBLambda"
       Description: An Application Load Balancer Lambda Target that returns regional metrics
       Handler: index.lambda_handler
-      Runtime: python3.7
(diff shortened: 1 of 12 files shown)


parikhudit:
@keithrozario Thanks for your contribution and creating this pull request. Unfortunately we cannot directly merge this as changes in each pattern usually goes through review testing as per best practices we follow.
It we be really helpful if you could create a pull request per pattern and (optionally) add test steps followed so it is easier for us to review, test and merge. Thanks again for the changes and look forward to you individual pull requests.

PR #2323, opened 2024-06-12
Navbalaraman feature api lambda sfn sagemaker

navbalaraman:
*Issue #, if available:* 2324

*Description of changes:*
New Pattern: API driven SageMaker domain creation


bfreiberg:
Hi @navbalaraman and thank you for your PR. Can you help me understand what the added benefit is over creating the domain directly via an Infrastructure-as-Code tool like CDK or Terraform?

navbalaraman:
Below are some of the benefits:
- these apis can be integrated with a portal allowing self-service capabilities to teams to provision/update/destroy the sagemaker based platform and its resources.
- these apis can be enhanced to support CRUD operations like update and delete sagemaker domain and its configurations.
- step functions workflow can be enhanced to add sagemaker persona roles and user profiles.
- reduces operational overhead since the platform management can fully automated.
- no additional deployment needed in case of updated like new domain creation/user profiles addition etc.

bfreiberg:
Thanks @navbalaraman for your reply. I'm not sure I understand the benefits. Please find my comments below

- these apis can be integrated with a portal allowing self-service capabilities to teams to provision/update/destroy the sagemaker based platform and its resources.
  - Isn't that true for the underlying boto3 calls as well?
- these apis can be enhanced to support CRUD operations like update and delete sagemaker domain and its configurations.
  - Wouldn't this still be just a wrapped around AWS APIs?
- step functions workflow can be enhanced to add sagemaker persona roles and user profiles.
- reduces operational overhead since the platform management can fully automated.
  - What is the platform in this regard?
- no additional deployment needed in case of updated like new domain creation/user profiles addition etc.
  - How would you handle the update of user profiles addition without a deployment?

navbalaraman:
- these apis can be integrated with a portal allowing self-service capabilities to teams to provision/update/destroy the sagemaker based platform and its resources.
Q: Isn't that true for the underlying boto3 calls as well?
A: the lambdas in the pattern use boto3 for AWS api calls. 
- these apis can be enhanced to support CRUD operations like update and delete sagemaker domain and its configurations.
Q: Wouldn't this still be just a wrapped around AWS APIs?
A: It could be one more AWS APIs that are handled in the lambdas/statemachine. The users of the API Gateway Rest API are not exposed to the complexity of the underlying domain management & configurations.
- step functions workflow can be enhanced to add sagemaker persona roles and user profiles.
- reduces operational overhead since the platform management can fully automated.
Q: What is the platform in this regard?
A: SageMaker domain and configurations constitute a platform which the data scientists are given access to use.
no additional deployment needed in case of updated like new domain creation/user profiles addition etc.
Q: How would you handle the update of user profiles addition without a deployment?
A: By accepting the users to be added in the api payload and adding another step in the state machine to process the users to create user profiles accordingly.

Note: This is a real use case that we have built and delivered to one of our customers who were looking for a fully self-service based portal for creating and managing infrastructure for data scientists to build, train and deploy models.

bfreiberg:
Thanks for your response. I see the value of having a managed solution to manage SageMaker resources without having needed access to the underlying infrastructure. However, I don't see this fitting the pattern definition. Normally, I would ask you to submit this as a repo to Serverlessland as it is more a full solution. Unfortunately, intake of repos is currently paused. 
Therefore, I would suggest to submit it to aws-samples as a standalone sample.

PR #2335, opened 2024-06-23
update `cdk-sfn-s3` pattern to cdk v2

garysassano:
*Issue #, if available:*

*Description of changes:*

Updated the code to use CDK v2 libs.


pputhran (changes requested):
can you also add example-pattern.json?

[example](https://github.com/aws-samples/serverless-patterns/blob/main/_pattern-model/example-pattern.json)

cdk-sfn-s3/example-pattern.json
```diff
@@ -0,0 +1,54 @@
...
+    "text": [
+      "Step Functions receives an execution request using the `start-execution` API command and puts an object to an S3 bucket using the `PutObject` API call from the AWS SDK service integration."
+    ]
+  },
+  "gitHub": {
+    "template": {
+      "repoURL": "https://github.com/aws-samples/serverless-patterns/tree/main/cdk-sfn-s3",
+      "templateURL": "serverless-patterns/cdk-sfn-s3",
+      "projectFolder": "cdk-sfn-s3",
+      "templateFile": "src/lib/cdk-sfn-s3-stack.ts"
```
pputhran, line 18: Running into an error on validating this file. Can you please remove the projectFolder value from the templateFile?

Simple fix would be to rename the `cdk-sfn-s3-stack.ts` to something else and update the `templateFile` value

garysassano:
How many months does it usually take for PRs to get merged?

cdk-sfn-s3/example-pattern.json
garysassano: I know it's been half a year, but could you please take a moment to review the PR?

parikhudit:
Thanks for the contribution @garysassano, appreciate your ownership in updating existing pattern.
Just a heads up that I've reverted the author name in ```example-pattern.json``` to be of initial author. We do not generally add new author to the pattern, unless there's a major revamp or the pattern is re-written.

PR #2352, opened 2024-07-05
New serverless pattern - appsync-private-api-sam

Tessot:
*Issue #, if available:*

*Description of changes:* This pattern shows how you can deploy an AWS AppSync Private API which can only be invoked by resources within your private network.


Tessot:
This contains two new patterns which is not recommended hence will have to close and create a separate pull request with only one new pattern added to it.

PR #2429, opened 2024-08-13
Adding Git Release

Vandita2020:
**Description:**
This PR is to support downloading the code for a pattern easily. By creating a release for each pattern code, it'll be easy to download the code for a user by just downloading the respective asset of that pattern.

This PR adds a new GitHub Action that will create a release (v1.0.0) with all pattern folders saved as asset in the release. If the PR already exists and there are new changes to the repo, it'll update the respective asset in the release.

**Acknowledgement**

### DIFF
--- a/.github/workflows/create_update_release.yaml
+++ b/.github/workflows/create_update_release.yaml
@@ -1,0 +1,143 @@
+name: Create and Update Releases
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  get-release:
+    runs-on: ubuntu-latest
+    outputs: 
+      latest_release: ${{ steps.get_latest_release.outputs.latest_release }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Get the latest release tag
+        id: get_latest_release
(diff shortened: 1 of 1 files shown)


.github/workflows/create_update_release.yaml
```diff
@@ -0,0 +1,143 @@
...
+            archive="temp_archives/$folder.zip"
+            if [ -f "$archive" ]; then
+              gh release upload ${{ needs.get-release.outputs.latest_release }} "$archive" --clobber
+              echo "Uploaded $archive, waiting to avoid rate limits..."
+            else
+              echo "Archive $archive not found"
+            fi
+          done < changed_folders.txt
+      
+      - name: Clean up
```
roger-zhangg, line 138: I'm a bit confused, I believe Github Actions uses ephemeral storage. Is there a specific reason we clean up here?

Vandita2020: You're right, GitHub Actions uses virtual machines that get disposed completely once the process is done. I added clean up here for good practice. This could be removed and it won't make any difference.

PR #2457, opened 2024-08-29
Updated policy so lambda function role has proper permissions

seshubaws:
*Issue #, if available:*

*Description of changes:*
Previous policy gave me the below error when trying to invoke the Lambda function:
```
errorType": "AmazonS3Exception",
  "errorMessage": "User: arn:aws:sts::683517028648:assumed-role/s3dotnet-ResizerFunctionRole-TZwnuYpDb5Rv/s3dotnet-ResizerFunction-12oroDMZmSGE is not authorized to perform: s3:GetObject on resource: \"arn:aws:s3:::imageresizer-683517028648/seshub-photo.jpeg\" because no identity-based policy allows the s3:GetObject action"
```

Updated the policy so the Lambda function role has proper permissions. Tested that invocation works with updated policy.

### DIFF
--- a/s3-lambda-dotnet/ImageResize/Function.cs
+++ b/s3-lambda-dotnet/ImageResize/Function.cs
@@ -103,7 +103,7 @@ public class Function
                 }
 
                 LambdaLogger.Log("----> Thumbnail file Key: " + thumbnailObjectKey);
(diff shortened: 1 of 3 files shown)


julianwood:
Thanks

PR #2539, opened 2024-12-12
New serverless pattern - lambda-kendra-bedrock-servicenow

sritang0805:
*Issue #, if available:*

*Description of changes:*
Added ServiceNow Data connector integration for lambda kendrock bedrock pattern.


ellisms:
@sritang0805 Thank you for the submission. We already have several patterns for Kendra connectors and there is not a substantial difference between each connector configuration. Therefore, we don't see a need to have a pattern per connector.

PR #2546, opened 2024-12-18
apigw certificate bound access token example added

kbdraai:
*Issue #, if available:*

*Description of changes:*
Added certificate bound access token support using API Gateway and Cognito. Provided through a SAM template.


ellisms:
@kbdraai I started testing this, but the Lambda Authorizer function fails with the following error:
```
[ERROR] Runtime.ImportModuleError: Unable to import module 'handlers.authorizer': PyO3 modules compiled for CPython 3.8 or older may only be initialized once per interpreter process
Traceback (most recent call last):
```

kbdraai:
@ellisms  Did you run a sam build with use_container=true? 
I ran through it again and it works for me. The issue seems with the build phase as it seems that it was build with python 3.8 modules. 

Can you elaborate on the process you followed? 
The OS?
Did you fill in the complete the samconfig.toml file with the values required?

kbdraai:
This looks like an issue with the token. To make it easier to get the token I used the AdminInitiateAuth Cognito API:

```
class CognitoAuth(AuthBase):
    def __init__(self, user_pool_id, client_id, username, password):
        self.user_pool_id = user_pool_id
        self.client_id = client_id
        self.username = username
        self.password = password
        self.token = self.authenticate_user()

    def authenticate_user(self):
        client = boto3.client('cognito-idp', region_name='us-east-1')
        try:
            response = client.admin_initiate_auth(
                UserPoolId=self.user_pool_id,
                ClientId=self.client_id,
                AuthFlow='ADMIN_USER_PASSWORD_AUTH',
                AuthParameters={
                    'USERNAME': self.username,
                    'PASSWORD': self.password,
                }
            )
            return response['AuthenticationResult']['AccessToken']
        except client.exceptions.NotAuthorizedException:
            raise Exception("The username or password is incorrect")
        except Exception as e:
            raise Exception(f"An error occurred: {str(e)}")

    def __call__(self, r):
        r.headers['Authorization'] = f'Bearer {self.token}'
        return r
```

You can also manually decode the token to ensure that it is correct.

apigw-cognito-certificate-bound-access-token/README.md
```diff
@@ -0,0 +1,89 @@
...
+openssl x509 -in client-cert.crt -noout -fingerprint -sha256 | cut -d'=' -f2 | tr -d ':' | xxd -r -p | base64 | tr -d '='
+```
+
+2. Navigate to the Cognito User Pool created in the solution. [Then create and verify](https://docs.aws.amazon.com/cognito/latest/developerguide/signing-up-users-in-your-app.html) a user.
+
+3. Add the certificate fingerprint to the `custom:cert_fingerprint` custom attribute of the user.
+
+4. [Create a DNS record](https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/routing-to-api-gateway.html) for the custom domain.
+
+5. [Get an access token](https://docs.aws.amazon.com/cognito/latest/developerguide/authentication-flows-public-server-side.html) from Cognito.
```
ellisms, line 68: I recommend providing instructions for using the cognito-idp CLI to retrieve the token. I think it will be easier.

apigw-cognito-certificate-bound-access-token/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+```
+curl -v https://<your_custom_domain>/example --cert client-cert.crt --key client-cert.key -H "Authorization: <certificate_bound_access_token>"
+```
+
+## Cleanup
+ 
+1. Delete the stack
+    ```bash
+    aws cloudformation delete-stack --stack-name STACK_NAME
```
ellisms, line 80: use `sam delete` since you are using SAM.

ellisms:
Thanks @kbdraai I was able to test this. Please review the requested changes and make the updates before we move forward.

kbdraai:
@ellisms ive updated README.MD. 

For the deploying the deploy command will be very bulky with all of the parameters which is why I opted for samconfig.toml. Let me know if this is still okay. I can make the change.

I included the same code to get an access token using boto3. Let me know if this is fine as well. Otherwise I can add AWS CLI command.

kbdraai:
Changes pushed

ellisms:
Thanks, adding this to the queue for publishing.

PR #2579, opened 2025-01-06
New serverless pattern: eventbridge-pipes-sqs-to-stepfunctions-terraform

nareshrajaram2017:
*Issue #, if available:*

*Description of changes:*

The purpose of this pattern is to demonstrate how to use Amazon EventBridge Pipes to connect an Amazon SQS queue with an AWS Step Functions state machine. The pipe applies a filter and transformation before triggering the Step Functions workflow.
In this architecture, EventBridge Pipes receives messages from the SQS queue, filters them based on a specific condition, transforms the message format, and then invokes the Step Functions state machine.
The pipe will filter messages where the status is "COMPLETE" and transform the message structure before sending it to Step Functions.


bfreiberg:
Hi, thanks for creating this request. However, the foldername `eventbridge-pipes-sqs-to-stepfunctions-terraform` already exists. Could you please what the difference to the existing pattern is? Thanks

nareshrajaram2017:
Apologies, please disregard this. I will work on a different use case. Thanks for catching this.

PR #2588, opened 2025-01-15
New Serverless pattern - multi-account centralized API Gateway with Transit Gateway

j-liburd:
This is the initial commit for the project.


biswanathmukherjee:
@j-liburd This needs to be submitted as an application rather than pattern considering the volume of code changes involved.  Hence I am closing the PR. We are now accepting patterns with service to service connectivity with small amount of custom codes. However, you can submit this as an application.

PR #2597, opened 2025-01-22
New serverless pattern - apigw-lambda-transcribe-sam

Avs163:
Issue #, if available:

Description of changes:

This pattern enables serverless audio transcription by processing audio files through Amazon Transcribe. When files are uploaded to S3, API Gateway receives requests and triggers Lambda functions, which initiate transcription jobs using Amazon Transcribe, converting spoken content into searchable, timestamped text that can be used for subtitling, content analysis, or speech analytics.


ellisms:
Hi @Avs163 Thanks for the submission. Upon reviewing the pattern, it is very similar to an [existing pattern](https://serverlessland.com/patterns/s3-lambda-transcribe-sam), except APIGW integration. Given S3 integrates directly with Lambda, I don't think adding APIGW adds value here. I'm going to close this without merging. Feel free to continue to submit other patterns.

PR #2661, opened 2025-02-04
Add new pattern: API Gateway to SQS with Priority-Based Routing

kushalgm:
This pattern demonstrates how to implement priority-based message routing from API Gateway directly to multiple SQS queues without using Lambda functions.

Key features:
- Direct integration between API Gateway and SQS using VTL mapping templates
- Dynamic routing to three different queues (high, default, low) based on a priority query parameter
- Secure IAM role configuration for API Gateway to SQS permissions
- Complete CloudFormation template for infrastructure as code

Example usage:
- High priority messages: /endpoint?priority=high
- Default priority: /endpoint (no parameter needed)
- Low priority messages: /endpoint?priority=low

The pattern includes:
- Detailed testing instructions with curl commands
- Example events and expected responses
- Clean architecture that follows AWS best practices

### DIFF
--- a/apigw-priority-sqs-routing/README.md
+++ b/apigw-priority-sqs-routing/README.md
@@ -1,0 +1,99 @@
+# API Gateway to SQS with Priority-Based Message Routing
+
+This pattern demonstrates how to implement direct integration between API Gateway and multiple SQS queues using mapping templates, enabling dynamic message routing based on request parameters without using Lambda functions.
+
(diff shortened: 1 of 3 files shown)


marakere:
@kushalgm - Thank you for taking time to contribute to ServerlessLand.
As discussed, this pattern matches one that is already in ServerlessLand, so I am closing this request.

PR #2674, opened 2025-02-20
New Pattern: Cross-account APIs using Amazon API Gateway and AWS PrivateLink

stanlk:
Created SAM deployment for next pattern:
Cross-account APIs using Amazon API Gateway and AWS PrivateLink


ellisms:
Hi @stanlk - thanks for the submission. I think this solution is too complex to showcase as a pattern on this site. I think it is a better fit for aws-samples serverless-samples and encourage you to submit there: https://github.com/aws-samples/serverless-samples

PR #2706, opened 2025-04-05
New serverless pattern - apigw-private-cdn-private-ca

tustha:
**Amazon API Gateway private REST API with private custom domain name configured with private SSL certificate managed my ACM signed by Amazon Private Certificate Authority** 
This pattern enables secure access to a private REST API Gateway using a private custom domain name. The solution utilizes SSL certificates managed by AWS Certificate Manager (ACM) and signed by Amazon Private Certificate Authority (PCA), ensuring secure and authenticated communication within the private network.

PR #2708, opened 2025-04-05
New serverless pattern - apigw-private-cdn-acm

tustha:
**Amazon API gateway private REST API with private custom domain name configured with private SSL cert imported and managed by ACM**
This pattern enables secure access to a private REST API Gateway using a private custom domain name. The solution utilizes SSL certificates, imported and managed through AWS Certificate Manager (ACM).


vsr2158 (commented):
Should we change :
"Create a private REST API in API gateway" to
"Create a private REST API in API gateway with custom domain name"

Change:
"Deploys a private REST API through API Gateway" to
"Deploy a private REST API through API Gateway"

Is the step "cd serverless-patterns/apigw-private-cdn" correct? should it be apigw-private-cdn-acm?

Does the S3 bucket need to be in the same region as the deployment? if yes up[date the Requirement section.

bfreiberg:
Hi @tustha, thanks for your recent pattern submission. We appreciate your effort. Looking at this and #2706, we feel that these are very similar and could be combined into one. I would suggest to keep #2706  and mention the ability to import certificates in the README.

PR #2710, opened 2025-04-07
adding new pattern : step functions to on-premises API (terraform)

jeromevdl:
*Issue #, if available:*

*Description of changes:* Adding a new pattern to show how to call an on-premises API from Step Functions, using Terraform.


.gitignore
```diff
@@ -206,11 +206,12 @@ crash.log
...
 
 # Exclude all .tfvars files, which are likely to contain sensitive data, such as
-# password, private keys, and other secrets. These should not be part of version 
-# control as they are data points which are potentially sensitive and subject 
+# password, private keys, and other secrets. These should not be part of version
+# control as they are data points which are potentially sensitive and subject
 # to change depending on the environment.
 *.tfvars
 *.tfvars.json
+!stepfunctions-eventbridge-onpremise-tf/example.tfvars
```
parikhudit: Instead of creating `example.tfvars` adding pattern specific statement in parent `.gitignore`, can we rather just give an example of `example.tfvars` in pattern's README?

parikhudit: @jeromevdl can you take a look? ^

PR #2742, opened 2025-05-26
Pattern update - apigw-lambda-bedrock-sam

Sliiiin:
I've updated the bedrock_integration.py file to work with Claude 3.5 using the messaging API. Here are the changes I made:

1. Changed the request format from the older completion API to the messaging API:
   • Replaced the prompt field with a structured messages array
   • Added the required anthropic_version parameter set to "bedrock-2023-05-31"
   • Changed max_tokens_to_sample to max_tokens
   • Removed the stop_sequences parameter as it's handled differently in the messaging API

2. Updated the response handling:
   • The messaging API returns a different response structure with content in an array
   • Added code to extract the completion text from response_body.get('content', [{}])[0].get('text', '')
   • Extracted the stop reason from response_body.get('stop_reason', '')
   • Formatted the response to maintain backward compatibility with your existing API

3. Updated README slightly to reflect the changes.


Sliiiin:
Lambda layer is not required and was removed. README references to the Lambda Layer have also been removed.

bfreiberg:
Looks good, thank you for your contribution.

PR #2788, opened 2025-07-12
terraform-eventbridge-scheduled-lambda: Update runtime to python3.13

kakakakakku:
*Issue #, if available:*
N/A

*Description of changes:*

To prevent future deployment issues, I updated the Lambda Python runtime version to ~~`python3.12`~~`python3.13`.

While testing `terraform-eventbridge-scheduled-lambda`, I noticed that the Lambda runtime version `python3.8` was deprecated. Although it's still deployable at the moment, it will not be allowed after **October 1, 2025**.

## Check

> [!WARNING]
> Warning is likely caused by the fact that the AWS provider version is currently at v6 (due to `>= 4.9` constraint), and `data.aws_region.current.name` is now a deprecated attribute. There is no functional impact at this time👍 

```
╷
│ Warning: Deprecated attribute
│ 
│   on .terraform/modules/lambda_function/outputs.tf line 9, in output "lambda_function_arn_static":
│    9:   value       = local.create && var.create_function && !var.create_layer ? "arn:aws:lambda:${data.aws_region.current.name}:${data.aws_caller_identity.current.account_id}:function:${var.function_name}" : ""
│ 
│ The attribute "name" is deprecated. Refer to the provider documentation for details.
│ 
│ (and one more similar warning elsewhere)
╵
```

image

Thank you😀

---

### DIFF
--- a/apigw-sqs-terraform/main.tf
+++ b/apigw-sqs-terraform/main.tf
@@ -10,7 +10,6 @@ terraform {
 }
 
 provider "aws" {
-  profile = "default"
   region  = "us-east-1"
 }
 


apigw-sqs-terraform/main.tf
```diff
@@ -10,7 +10,6 @@ terraform {
 }
 
 provider "aws" {
-  profile = "default"
```
kakakakakku, line 13: note: I think it’s better if the user can choose the profile to use😀 For example, I also referred to **terraform-s3-lambda** and **terraform-lambda-sfn**.
- https://github.com/aws-samples/serverless-patterns/blob/main/terraform-s3-lambda/main.tf
- https://github.com/aws-samples/serverless-patterns/blob/main/terraform-lambda-sfn/main.tf

terraform-eventbridge-scheduled-lambda/main.tf
```diff
@@ -43,7 +43,7 @@ module "lambda_function" {
   function_name = "${random_pet.this.id}-lambda"
   description   = "My awesome lambda function"
   handler       = "index.lambda_handler"
-  runtime       = "python3.8"
+  runtime       = "python3.12"
```
biswanathmukherjee: Better to update it to latest supported versopm 3.13.

kakakakakku: @biswanathmukherjee 
Thank you for your review!
OK. I updated to Python 3.13 and fetch latest HEAD branch for easier merging😀

PR #2817, opened 2025-08-22
dynamodb-streams-lambda-terraform: Update runtime to nodejs22.x

kakakakakku:
*Issue #, if available:*
N/A

*Description of changes:*
To prevent future deployment issues, I updated the Lambda Node.js runtime version to `nodejs22.x`.

While testing `dynamodb-streams-lambda-terraform`, I noticed that the Lambda runtime version `nodejs16.x` was deprecated. Although it's still deployable at the moment, it will not be allowed after **October 1, 2025**.

## Check

image

Thank you😀

---

### DIFF
--- a/dynamodb-streams-lambda-terraform/README.md
+++ b/dynamodb-streams-lambda-terraform/README.md
@@ -61,7 +61,7 @@ After deployment, add an item to the DynamoDB table. Go to the CloudWatch Logs f
     ```bash
     terraform show
     ```
-    ```
+
 ----
 Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 


dynamodb-streams-lambda-terraform/src/index.js
kakakakakku, line 1: note: I renamed the file because it had the following error.

```
╷
│ Error: Archive creation error
│ 
│   with data.archive_file.lambda_zip_file,
│   on main.tf line 46, in data "archive_file" "lambda_zip_file":
│   46: data "archive_file" "lambda_zip_file" {
│ 
│ error creating archive: error archiving file: could not archive missing file: ./src/index.js
╵
```

dynamodb-streams-lambda-terraform/main.tf
```diff
@@ -2,7 +2,7 @@ terraform {
   required_providers {
     aws = {
       source  = "hashicorp/aws"
-      version = "~> 3.27"
+      version = "~> 5.0"
```
kakakakakku, line 5: note: Older version does not support Node.js 22 runtime. So I updated the version to v5.

PR #2818, opened 2025-08-22
lambda-function-url-terraform: Update runtime to nodejs22.x

kakakakakku:
*Issue #, if available:*
N/A

*Description of changes:*

To prevent future deployment issues, I updated the Lambda Node.js runtime version to `nodejs22.x`.

While testing `lambda-function-url-terraform`, I noticed that the Lambda runtime version `nodejs16.x` was deprecated. Although it's still deployable at the moment, it will not be allowed after **October 1, 2025**.

## Check

```sh
$ awscurl --service lambda https://q3hq2fhxhp6ibudvwdbgq7vkay0mwixd.lambda-url.us-east-1.on.aws/
Hello World
```

Thank you😀

---

### DIFF
--- a/lambda-function-url-terraform/.gitignore
+++ b/lambda-function-url-terraform/.gitignore
@@ -1,0 +1,1 @@
+lambda.zip


lambda-function-url-terraform/main.tf
```diff
@@ -2,7 +2,7 @@ terraform {
   required_providers {
     aws = {
       source  = "hashicorp/aws"
-      version = "~> 4.21"
+      version = "~> 5.0"
```
kakakakakku, line 5: note: Older version does not support Node.js 22 runtime. So I updated the version to v5.

lambda-function-url-terraform/main.tf
```diff
@@ -54,6 +53,11 @@ resource "aws_iam_role" "lambda_iam_role" {
 EOF
 }
 
+resource "aws_iam_role_policy_attachment" "lambda_basic_execution" {
+  role       = aws_iam_role.lambda_iam_role.name
+  policy_arn = data.aws_iam_policy.lambda_basic_execution_role_policy.arn
+}
```
kakakakakku, line 59: note: The `managed_policy_arns` argument is already deprecated. I update code correctly😀 See document below.

- https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/iam_role

PR #2967, opened 2026-02-19
adding nodejs webhook df pattern

ginjups:
*Issue #, if available:*

*Description of changes:*

adding nodejs webhook df pattern

### DIFF
--- a/lambda-durable-webhook-sam-nodejs/README.md
+++ b/lambda-durable-webhook-sam-nodejs/README.md
@@ -39,8 +39,6 @@ This pattern acheives the following key features:
 
 **Note:** Each step writes status updates to DynamoDB before its main work. These writes are idempotent, so retries are safe. During replay, the DynamoDB status reflects the last successfully written state—not the current replay position. Status queries should treat intermediate states as "in progress."
 
-## Important
-
 **Important:** Please check the [AWS documentation](https://docs.aws.amazon.com/lambda/latest/dg/durable-functions.html) for regions currently supported by AWS Lambda durable functions.
 
 ## Deployment


lambda-durable-webhook-sam-nodejs/README.md
marcojahn, line 1: Can you please align your README.md with the one listed here: https://github.com/aws-samples/serverless-patterns/tree/main/_pattern-model

- Order of items
- How It Works, Webhook Processing Workflow (3 Steps), Key Features can be condensed in one item
- Required IAM Permissions - can be removed
- Components section can be removed entirely
- API Endpoints, Monitoring, Error Handling, Cost Optimization, Security Considerations, NodeJS Implementation Notes can be removed
- Learn More can be integrated into the first description using Links to the docs
- The README does not mention that deployed resources may incur AWS charges. While the pattern uses pay-per-use services, a cost warning helps users understand they should clean up after testing.
- The README does not mention that deployed resources may incur AWS charges. While the pattern uses pay-per-use services, a cost warning helps users understand they should clean up after testing.

If you like, you
- can add numbers to your architecture diagram and move/re-use content blocks (please as short/crisp as possible) to the architecture diagram (below) in a written form.

lambda-durable-webhook-sam-nodejs/README.md
```diff
@@ -0,0 +1,202 @@
...
+- **CloudWatch Logs**: `logs:DescribeLogGroups`, `logs:FilterLogEvents`, `logs:GetLogEvents`, `logs:TailLogEvents`
+
+## Deployment
+
+1. **Build the application**:
+   ```bash
+   sam build
+   ```
+
+2. **Deploy to AWS**:
```
marcojahn, line 51: Please note required inputs `WebhookSecret`

lambda-durable-webhook-sam-nodejs/README.md
```diff
@@ -0,0 +1,202 @@
...
+   ```bash
+   sam deploy --guided
+   ```
+   
+   Note the outputs after deployment:
+   - `WebhookApiUrl`: Use this for sending webhook POST requests
+   - `StatusQueryApiUrl`: Use this for querying execution status
+
+3. **Test the webhook**:
```
marcojahn:
```suggestion
## Testing
```
Own section please

lambda-durable-webhook-sam-nodejs/README.md
```diff
@@ -0,0 +1,202 @@
...
+     }'
+   ```
+
+4. **Query webhook status**:
+   ```bash
+   # Get execution status (use executionToken from webhook response)
+   curl <StatusQueryApiUrl>
+   ```
+   
+   **Success indicators:**
```
marcojahn, line 83: Please ad a way testing the failure handling & timeout (set to a low limit) as it is described as a key capability of the sample

lambda-durable-webhook-sam-nodejs/template.yaml
```diff
@@ -0,0 +1,247 @@
...
+      DurableConfig:
+        ExecutionTimeout: 3600
+        RetentionPeriodInDays: 7
+      Policies:
+        Statement:
+          - Effect: Allow
+            Action:
+              - lambda:CheckpointDurableExecutions
+              - lambda:GetDurableExecutionState
+            Resource: !Sub 'arn:aws:lambda:${AWS::Region}:${AWS::AccountId}:function:${Environment}-webhook-processor'
```
marcojahn:
```suggestion
              - AWSLambdaBasicExecutionRole
              - AWSLambdaBasicDurableExecutionRolePolicy
```
Should be fine to be replaced with managed policies

lambda-durable-webhook-sam-nodejs/README.md
```diff
@@ -0,0 +1,202 @@
...
+- ✅ **Failure Recovery** - Resumes from last checkpoint on failure
+- ✅ **Asynchronous Processing** - Immediate 202 response, processing in background
+- ✅ **State Persistence** - Execution state stored in DynamoDB with TTL
+- ✅ **Status Query API** - Real-time status tracking via REST API
+
+## Important
+
+⚠️ **Important:** Please check the [AWS documentation](https://docs.aws.amazon.com/lambda/latest/dg/durable-functions.html) for regions currently supported by AWS Lambda durable functions.
+
+Learn more about this pattern at Serverless Land Patterns: https://serverlessland.com/patterns/lambda-durable-webhook-sam-python
```
marcojahn:
```suggestion
Learn more about this pattern at Serverless Land Patterns: [https://serverlessland.com/patterns/lambda-durable-webhook-sam-nodejs](https://serverlessland.com/patterns/lambda-durable-webhook-sam-nodejs)
```
should be the nodejs pattern, not the python one

lambda-durable-webhook-sam-nodejs/README.md
```diff
@@ -0,0 +1,202 @@
...
+
+## Prerequisites
+
+- [AWS CLI v2](https://docs.aws.amazon.com/cli/latest/userguide/install-cliv2.html) installed and configured
+- [AWS SAM CLI](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/serverless-sam-cli-install.html) installed
+- [Node.js 24.x](https://nodejs.org/en/download/) runtime installed
+- [Docker](https://docs.docker.com/get-docker/) (for containerized builds)
```
marcojahn: Docker really needed?

lambda-durable-webhook-sam-nodejs/src/webhook_validator/index.js
marcojahn, line 1: The webhook_processor function uses ES modules (import/export with "type": "module" in package.json), but webhook_validator and status_query use CommonJS (exports.handler, require()). Since the SAM template sets a global runtime of nodejs24.x, all functions should use a consistent module system. The inconsistency will confuse readers.

lambda-durable-webhook-sam-nodejs/template.yaml
```diff
@@ -0,0 +1,247 @@
...
+            Resource: 
+              - !GetAtt WebhookValidatorFunction.Arn
+          - Effect: Allow
+            Action:
+              - dynamodb:GetItem
+              - dynamodb:PutItem
+              - dynamodb:UpdateItem
+              - dynamodb:DeleteItem
+              - dynamodb:Query
+              - dynamodb:Scan
```
marcojahn, line 83: The WebhookProcessorFunction IAM policy includes dynamodb:Scan permission, but the application code only uses PutCommand, UpdateCommand, and GetCommand. Scan is an expensive operation that reads every item in the table and should not be granted unless explicitly needed.

lambda-durable-webhook-sam-nodejs/template.yaml
```diff
@@ -0,0 +1,247 @@
+AWSTemplateFormatVersion: '2010-09-09'
+Transform: AWS::Serverless-2016-10-31
+Description: 'Webhook Receiver Pattern using AWS Lambda durable functions with Python - NodeJS version'
```
marcojahn, line 3: Can you pls fix the description (python)

lambda-durable-webhook-sam-nodejs/template.yaml
```diff
@@ -0,0 +1,247 @@
...
+          AttributeType: S
+        - AttributeName: timestamp
+          AttributeType: N
+      KeySchema:
+        - AttributeName: executionToken
+          KeyType: HASH
+      TimeToLiveSpecification:
+        AttributeName: ttl
+        Enabled: true
+      GlobalSecondaryIndexes:
```
marcojahn, line 40: Is the GSI used anywhere? Did not find a reference.

All three Lambda functions only perform:

PutCommand - writes new items (webhook_processor)
UpdateCommand - updates existing items by primary key (webhook_processor)
GetCommand - retrieves items by primary key executionToken (status_query)
None of them use DynamoDB QueryCommand or ScanCommand operations that would leverage the GSI

**And/Or**

he `TimestampIndex` GSI uses `ProjectionType: ALL`, which duplicates all attributes into the index. The GSI's partition key is `timestamp` (a number), which is not a good partition key for DynamoDB as it provides no meaningful access pattern. If the GSI is intended for time-based queries, consider whether it is actually needed or if a different key design would be more appropriate.

lambda-durable-webhook-sam-nodejs/src/webhook_processor/index.js
```diff
@@ -0,0 +1,230 @@
...
+        
+        // Extract configuration from environment
+        const eventsTableName = process.env.EVENTS_TABLE_NAME;
+        const environment = process.env.ENVIRONMENT || 'dev';
+        
+        // Parse the incoming webhook event
+        const webhookPayload = JSON.parse(event.body || '{}');
+        
+        // Use executionToken from API Gateway or generate new one
+        const executionToken = event.executionToken || randomUUID();
```
marcojahn: When event.executionToken is not provided, randomUUID() runs outside any step. On replay, this code runs again and generates a different UUID. The executionToken is then used as the DynamoDB partition key in every subsequent step. This means:

First invocation: executionToken = "abc-123" → Steps 1-2 complete, checkpoint saved
Replay (after interruption): executionToken = "xyz-789" (new UUID!) → Step 1 returns cached result that references "abc-123", but all DynamoDB operations in Steps 2-3 now target "xyz-789" — a different record
The result: orphaned DynamoDB records, status queries returning stale data, and the workflow operating on the wrong execution state.

AWS documentation explicitly states:

"Non-deterministic operations, such as random number generation and UUIDs [...] should be wrapped in steps."

lambda-durable-webhook-sam-nodejs/src/webhook_processor/index.js
```diff
@@ -0,0 +1,230 @@
...
+        
+        console.log(`Processing webhook with execution token: ${executionToken}`);
+        console.log(`Webhook payload:`, JSON.stringify(webhookPayload, null, 2));
+        
+        try {
+            // Store initial execution state
+            await dynamodb.send(new PutCommand({
```
marcojahn, line 43: This DynamoDB write happens outside any step. On every replay, this PutCommand executes again, overwriting the record with status: 'STARTED' — even if the execution has already progressed to VALIDATING, PROCESSING, or COMPLETED. This resets the execution state visible to the status query API.

Additionally, Date.now() is non-deterministic and will produce different timestamps on each replay.

lambda-durable-webhook-sam-nodejs/src/webhook_processor/index.js
```diff
@@ -0,0 +1,230 @@
...
+        const environment = process.env.ENVIRONMENT || 'dev';
+        
+        // Parse the incoming webhook event
+        const webhookPayload = JSON.parse(event.body || '{}');
+        
+        // Use executionToken from API Gateway or generate new one
+        const executionToken = event.executionToken || randomUUID();
+        
+        console.log(`Processing webhook with execution token: ${executionToken}`);
+        console.log(`Webhook payload:`, JSON.stringify(webhookPayload, null, 2));
```
marcojahn: While console.log itself doesn't break determinism, these lines reference executionToken which (per DUR-1) is non-deterministic. On replay, these logs will print a different execution token than the one used in the checkpointed steps, making debugging extremely confusing.

More importantly, the durable SDK provides stepContext.logger inside steps (which the code already uses correctly within steps). Code outside steps should use minimal logging, and any logging that references execution-specific data should happen inside steps.

lambda-durable-webhook-sam-nodejs/src/webhook_processor/index.js
marcojahn, line 1:
```
// ❌ CURRENT — No name argument provided
const validationResult = await context.step(async (stepContext) => { ... });
const processingResult = await context.step(async (stepContext) => { ... });
const finalResult = await context.step(async (stepContext) => { ... });
```
```
// ✅ FIX
const validationResult = await context.step('validate-webhook', async (stepContext) => { ... });
const processingResult = await context.step('process-webhook', async (stepContext) => { ... });
const finalResult = await context.step('finalize-webhook', async (stepContext) => { ... });
```
AWS best practices state:

"Use descriptive names — Names like validate-order instead of step1 make logs and errors easier to understand"

marcojahn, line 1: Please review the following assumptions/findings and adept if necessary

```
// Lines 130 and 152 inside steps
processedAt: new Date().toISOString(),
':completedAt': new Date().toISOString()
```

These timestamps are inside steps, so they ARE checkpointed — the value is captured on first execution and replayed from cache on subsequent replays. This is technically correct for the checkpoint/replay model.

However, the issue is that these timestamps are written to DynamoDB within the same step. If the step retries (at-least-once semantics), the DynamoDB write executes again with a new timestamp, but the step's return value still contains the original timestamp from the first attempt. This creates an inconsistency between what's in DynamoDB and what's in the checkpoint.

```
// ✅ FIX — Generate timestamp once at the start of the step and use it consistently
const finalResult = await context.step('finalize-webhook', async (stepContext) => {
    const completedAt = new Date().toISOString();
    
    await dynamodb.send(new UpdateCommand({
        // ... use completedAt consistently
        ExpressionAttributeValues: {
            ':completedAt': completedAt
        }
    }));

    return {
        executionToken,
        status: "completed",
        completedAt  // Same value as written to DynamoDB
    };
});
```
This is acceptable for this pattern since the DynamoDB writes are idempotent status updates, but it's worth noting as a best practice.

lambda-durable-webhook-sam-nodejs/src/webhook_processor/index.js
```diff
@@ -0,0 +1,230 @@
...
+                        '#step': 'currentStep'
+                    },
+                    ExpressionAttributeValues: {
+                        ':status': 'VALIDATING',
+                        ':step': 'validate'
+                    }
+                }));
+
+                // Call the separate webhook validator function
+                const { LambdaClient, InvokeCommand } = await import('@aws-sdk/client-lambda');
```
marcojahn: Two issues here:

Dynamic import inside a step: The await import() is an async operation that adds latency on every first execution of this step. If the step retries, the import runs again. SDK clients should be initialized in the module scope (outside the handler) for connection reuse across warm invocations.

New client on every step execution: Creating a new LambdaClient inside the step means no HTTP connection reuse. The DynamoDB client is correctly initialized at module scope (lines 7-8), but the Lambda client is not.

```
// ✅ FIX — Move to module scope alongside DynamoDB client
import { LambdaClient, InvokeCommand } from '@aws-sdk/client-lambda';

const lambdaClient = new LambdaClient({});

```

lambda-durable-webhook-sam-nodejs/src/webhook_processor/index.js
```diff
@@ -0,0 +1,230 @@
...
+                const lambdaClient = new LambdaClient({});
+                
+                const validatorFunctionArn = process.env.WEBHOOK_VALIDATOR_FUNCTION_ARN;
+                const invokeResponse = await lambdaClient.send(new InvokeCommand({
+                    FunctionName: validatorFunctionArn,
+                    Payload: JSON.stringify({
+                        payload: webhookPayload,
+                        executionToken: executionToken
+                    })
+                }));
```
marcojahn, line 91: The durable execution SDK provides context.invoke() specifically for calling other Lambda functions from within a durable function. Using context.invoke() instead of manual SDK invocation provides:

Automatic checkpointing of the invocation result
Built-in retry handling
Proper integration with the durable execution lifecycle
The current approach wraps a manual Lambda invocation inside a step, which works but misses the SDK's built-in invocation support.

```
// ✅ FIX — Use context.invoke() for Lambda-to-Lambda calls
const validatorResult = await context.invoke(validatorFunctionArn, {
    payload: webhookPayload,
    executionToken: executionToken
});
```

lambda-durable-webhook-sam-nodejs/src/webhook_processor/index.js
```diff
@@ -0,0 +1,230 @@
...
+                UpdateExpression: 'SET #status = :status, #error = :error',
+                ExpressionAttributeNames: {
+                    '#status': 'status',
+                    '#error': 'error'
+                },
+                ExpressionAttributeValues: {
+                    ':status': 'FAILED',
+                    ':error': error.message
+                }
+            }));
```
marcojahn: The catch block is outside any step. If an error occurs and the function retries, the catch block's DynamoDB write runs on every retry attempt. More critically, if the error is transient and the function eventually succeeds on retry, the catch block from the failed attempt has already written FAILED status to DynamoDB — but the successful retry overwrites it. This creates a race condition in the status visible to the query API.

The durable SDK handles retries automatically for steps. The outer try/catch should only handle truly unrecoverable errors, and the error state write should be wrapped in a step.

```
// ✅ FIX — Let the SDK handle retries; wrap error handling in a step
} catch (error) {
    const errorResult = await context.step('handle-error', async () => {
        await dynamodb.send(new UpdateCommand({
            TableName: eventsTableName,
            Key: { executionToken },
            UpdateExpression: 'SET #status = :status, #error = :error',
            ExpressionAttributeNames: { '#status': 'status', '#error': 'error' },
            ExpressionAttributeValues: { ':status': 'FAILED', ':error': error.message }
        }));
        return { status: 'FAILED', error: error.message };
    });
    
    return {
        statusCode: 500,
        body: JSON.stringify({ message: 'Webhook processing failed', ...errorResult })
    };
}
```

lambda-durable-webhook-sam-nodejs/src/webhook_processor/index.js
```diff
@@ -0,0 +1,230 @@
...
+                    originalPayload: webhookPayload,
+                    businessResult: `Processed webhook of type: ${webhookPayload.type || 'unknown'}`,
+                    dataTransformed: webhookPayload.data ? JSON.stringify(webhookPayload.data).toUpperCase() : null,
+                    processedAt: new Date().toISOString(),
+                    metadata: {
+                        processedBy: 'webhook-processor-nodejs',
+                        version: '1.0.0'
+                    }
+                };
+            });
```
marcojahn, line 167:
```suggestion
return {
    executionToken,
    status: "processed",
    payloadType: webhookPayload.type || 'unknown',
    processedAt: new Date().toISOString()
};
```

AWS best practices state:

"Keep state minimal — Store IDs and references, not full objects. [...] Use Amazon S3 or DynamoDB for large data, pass references in state."

The originalPayload is already stored in DynamoDB (from the initial PutCommand). Including it in the step return value means it's also serialized into the checkpoint. For large webhook payloads, this doubles storage and increases checkpoint costs. The processingResult is then also embedded in Step 3's DynamoDB write and return value, tripling the data.

lambda-durable-webhook-sam-nodejs/src/webhook_processor/index.js
marcojahn, line 1: Please review

Each step writes a status update to DynamoDB before doing its main work. If a step retries, the DynamoDB UpdateCommand runs again. While DynamoDB UpdateCommand is naturally idempotent for these status writes (writing the same value is a no-op), the pattern should be documented as intentionally idempotent.

More importantly, if Step 1 completes but Step 2 fails and the function replays, Step 1's DynamoDB write does NOT re-execute (it's checkpointed), but the DynamoDB record still shows the status from Step 1's original execution. This is correct behavior, but the pattern should be aware that the DynamoDB status reflects the last successfully written status, not necessarily the current replay position.

No code change needed, but this should be documented in the README as a design consideration.

lambda-durable-webhook-sam-nodejs/README.md
ginjups: Addressed

lambda-durable-webhook-sam-nodejs/README.md
ginjups: Done

lambda-durable-webhook-sam-nodejs/README.md
ginjups: Updated Readme

lambda-durable-webhook-sam-nodejs/README.md
ginjups: Updated Readme

lambda-durable-webhook-sam-nodejs/template.yaml
ginjups: changed to AWS Managed Policy

lambda-durable-webhook-sam-nodejs/src/webhook_processor/index.js
ginjups: Updated code

lambda-durable-webhook-sam-nodejs/src/webhook_processor/index.js
ginjups: Addressed

lambda-durable-webhook-sam-nodejs/src/webhook_processor/index.js
ginjups: Done

lambda-durable-webhook-sam-nodejs/src/webhook_processor/index.js
ginjups: Modified

lambda-durable-webhook-sam-nodejs/src/webhook_processor/index.js
ginjups: Noted, mentioned in Readme

PR #3003, opened 2026-03-26
New serverless pattern - Amazon CloudFront to Amazon Bedrock AgentCore Runtime

rr-on-gh:
*Issue #, if available:*

*Description of changes:*

This pattern deploys an Amazon CloudFront distribution in front of Amazon Bedrock AgentCore Runtime to proxy requests with OAuth 2.0 authentication. Using Amazon CloudFront in front of AgentCore Runtime provides features to for production traffic serving applications like global edge caching, DDoS protection, AWS WAF integration for IP rate limiting and geo-blocking, custom domain support with SSL/TLS.


marakere:
@rr-on-gh - thanks for putting this together. Nice to see AgentCore Runtime patterns coming in, and the multi-protocol coverage (A2A, HTTP, MCP) is a great touch.

One thing I'd love to understand better: what's the customer use case or scenario that motivated putting CloudFront in front of AgentCore Runtime specifically?

Since agent interactions are inherently dynamic and per-request, the edge caching that's core to CloudFront's value proposition wouldn't typically apply here. I'm curious whether the primary driver is:
- WAF/DDoS protection
- Custom domain consolidation
- Something else entirely

It'd be great to call that out more explicitly in the README so users understand when to reach for this pattern vs. alternatives like API Gateway or direct Runtime access with native OAuth. This will certainly help our builders.

rr-on-gh:
> @rr-on-gh - thanks for putting this together. Nice to see AgentCore Runtime patterns coming in, and the multi-protocol coverage (A2A, HTTP, MCP) is a great touch.
> 
> One thing I'd love to understand better: what's the customer use case or scenario that motivated putting CloudFront in front of AgentCore Runtime specifically?
> 
> Since agent interactions are inherently dynamic and per-request, the edge caching that's core to CloudFront's value proposition wouldn't typically apply here. I'm curious whether the primary driver is:
> 
> * WAF/DDoS protection
> * Custom domain consolidation
> * Something else entirely
> 
> It'd be great to call that out more explicitly in the README so users understand when to reach for this pattern vs. alternatives like API Gateway or direct Runtime access with native OAuth. This will certainly help our builders.

@marakere thanks for picking this up. 

You are correct, edge caching does not apply here - I have removed it. 

The others, as you pointed out, along with custom auth handling via options like Lambda@Edge are what I see customers using CF in front of AgentCore Runtime. So this pattern should help simplify implementing that.

PR #3099, opened 2026-05-06
New pattern - dynamodb-cross-account-replication-cdk

NithinChandranR-AWS:
## Description
First pattern for **Amazon DynamoDB cross-account Global Tables replication** (Q1 2026 launch).

## What it does
- Creates DynamoDB Global Table with replica in us-west-2
- Cross-account IAM role for secure read access from replica account
- Point-in-time recovery enabled

## Testing
Deployed and tested. Write to us-east-1, read from us-west-2 replica confirmed with sub-second replication.

---

### DIFF
--- a/dynamodb-cross-account-replication-cdk/README.md
+++ b/dynamodb-cross-account-replication-cdk/README.md
@@ -1,0 +1,60 @@
+# Amazon DynamoDB Cross-Account Replication with Global Tables
+
+This pattern deploys a DynamoDB Global Table with cross-account replication and an IAM role for secure cross-account read access.
+
+Learn more about this pattern at Serverless Land Patterns: https://serverlessland.com/patterns/dynamodb-cross-account-replication-cdk
+
+Important: this application uses various AWS services and there are costs associated with these services after the Free Tier usage - please see the [AWS Pricing page](https://aws.amazon.com/pricing/) for details.
+
+## Requirements
+
+* [AWS CLI](https://docs.aws.amazon.com/cli/latest/userguide/install-cliv2.html) installed and configured
+* [AWS CDK](https://docs.aws.amazon.com/cdk/latest/guide/cli.html) installed
+* [Node.js](https://nodejs.org/en/download/) installed
+* Two AWS accounts (source and replica)
+
+## Deployment Instructions
+
+1. Clone and navigate to the pattern:
+    ```
+    cd serverless-patterns/dynamodb-cross-account-replication-cdk
+    npm install
+    ```
+2. Deploy with the replica account ID:
+    ```
+    cdk deploy --parameters ReplicaAccountId=123456789012 --parameters ReplicaRegion=us-west-2
+    ```
+
+## How it works
+
+- A DynamoDB Global Table is created with a replica in the specified region
+- DynamoDB automatically replicates all writes to the replica with sub-second latency
+- A cross-account IAM role allows the replica account to assume and read from the table
+- Point-in-time recovery is enabled for data protection
+
+## Testing
+
+```bash
+# Write an item to the source table
+aws dynamodb put-item --table-name $(aws cloudformation describe-stacks \
+  --stack-name DynamodbCrossAccountReplicationStack \
+  --query 'Stacks[0].Outputs[?OutputKey==`TableName`].OutputValue' --output text) \
+  --item '{"PK":{"S":"user#123"},"SK":{"S":"profile"},"name":{"S":"test"}}'
+
+# Read from replica region (same account)
+aws dynamodb get-item --table-name <TableName> \
+  --key '{"PK":{"S":"user#123"},"SK":{"S":"profile"}}' \
+  --region us-west-2
+```
+
+## Cleanup
+
+```
+cdk destroy
+```
+
+---
+
+Copyright 2026 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+SPDX-License-Identifier: MIT-0
--- a/dynamodb-cross-account-replication-cdk/README.md
+++ b/dynamodb-cross-account-replication-cdk/README.md
@@ -71,10 +71,21 @@ aws dynamodb get-item --table-name <TableName> \
 # Cross-account read (from reader account, assuming the role)
 aws sts assume-role --role-arn <CrossAccountRoleArn> \
   --role-session-name reader-test
-# Use returned credentials to read from either region
+
+# Export the temporary credentials
+export AWS_ACCESS_KEY_ID=<AccessKeyId from above>
+export AWS_SECRET_ACCESS_KEY=<SecretAccessKey from above>
+export AWS_SESSION_TOKEN=<SessionToken from above>
+
+# Read from the replica region using the cross-account role
+aws dynamodb get-item --table-name <TableName> \
+  --key '{"PK":{"S":"user#123"},"SK":{"S":"profile"}}' \
+  --region us-west-2
 ```
 
 ## Cleanup
+
+> **⚠️ Warning:** `cdk destroy` with `RemovalPolicy.DESTROY` will permanently delete the table and all its data. Back up any important data before destroying.
 
 ```bash
 cdk destroy
--- a/dynamodb-cross-account-replication-cdk/README.md
+++ b/dynamodb-cross-account-replication-cdk/README.md
@@ -1,6 +1,6 @@
-# Amazon DynamoDB Cross-Account Replication with Global Tables
+# Amazon DynamoDB Global Tables with Cross-Account Read Access
 
-This pattern deploys a DynamoDB Global Table with cross-account replication and an IAM role for secure cross-account read access.
+This pattern deploys a DynamoDB Global Table (same-account, multi-region) and an IAM role that grants a separate AWS account read-only access to the table in both regions.
 
 Learn more about this pattern at Serverless Land Patterns: https://serverlessland.com/patterns/dynamodb-cross-account-replication-cdk
 
@@ -10,27 +10,48 @@ Important: this application uses various AWS services and there are costs associ
 
 * [AWS CLI](https://docs.aws.amazon.com/cli/latest/userguide/install-cliv2.html) installed and configured
 * [AWS CDK](https://docs.aws.amazon.com/cdk/latest/guide/cli.html) installed
-* [Node.js](https://nodejs.org/en/download/) installed
-* Two AWS accounts (source and replica)
+* [Node.js](https://nodejs.org/en/download/) 20.x or later
+* Two AWS accounts (one owns the table, the other reads from it)
+
+## How it works
+
+```
+┌─────────────────────────────────────────────────────────────────┐
+│  Owner Account                                                   │
+│                                                                   │
+│  ┌─────────────────┐   Global Tables    ┌─────────────────┐     │
+│  │ DynamoDB Table   │ ─────────────────▶ │ DynamoDB Replica │     │
+│  │ (us-east-1)      │   auto-replicate   │ (us-west-2)      │     │
+│  └─────────────────┘                    └─────────────────┘     │
+│           │                                       │              │
+│           └───────────── IAM Role ───────────────┘              │
+│                              │                                    │
+└──────────────────────────────┼────────────────────────────────────┘
+                               │ AssumeRole
+┌──────────────────────────────┼────────────────────────────────────┐
+│  Reader Account              ▼                                    │
+│              GetItem / Query / Scan                               │
+└───────────────────────────────────────────────────────────────────┘
+```
+
+- A DynamoDB Global Table is created with a replica in the specified region (same account)
+- An IAM role is created that can be assumed by the reader account
+- The role grants read-only access (`GetItem`, `Query`, `Scan`, `BatchGetItem`) to the table in **both** regions
+- DynamoDB handles replication automatically with sub-second latency
+
+**Note:** This is same-account multi-region replication (Global Tables) with cross-account read access via IAM. For true cross-account Global Tables replication, use `TableV2MultiAccountReplica` in a separate stack deployed to the replica account.
 
 ## Deployment Instructions
 
 1. Clone and navigate to the pattern:
-    ```
+    ```bash
     cd serverless-patterns/dynamodb-cross-account-replication-cdk
     npm install
     ```
-2. Deploy with the replica account ID:
+2. Deploy with the reader account ID and replica region:
+    ```bash
+    cdk deploy --parameters ReplicaAccountId=123456789012 -c replicaRegion=us-west-2
     ```
-    cdk deploy --parameters ReplicaAccountId=123456789012 --parameters ReplicaRegion=us-west-2
-    ```
-
-## How it works
-
-- A DynamoDB Global Table is created with a replica in the specified region
-- DynamoDB automatically replicates all writes to the replica with sub-second latency
-- A cross-account IAM role allows the replica account to assume and read from the table
-- Point-in-time recovery is enabled for data protection
 
 ## Testing
 
@@ -41,15 +62,20 @@ aws dynamodb put-item --table-name $(aws cloudformation describe-stacks \
   --query 'Stacks[0].Outputs[?OutputKey==`TableName`].OutputValue' --output text) \
   --item '{"PK":{"S":"user#123"},"SK":{"S":"profile"},"name":{"S":"test"}}'
 
-# Read from replica region (same account)
+# Read from replica region (same account, verifies replication)
 aws dynamodb get-item --table-name <TableName> \
   --key '{"PK":{"S":"user#123"},"SK":{"S":"profile"}}' \
   --region us-west-2
+
+# Cross-account read (from reader account, assuming the role)
+aws sts assume-role --role-arn <CrossAccountRoleArn> \
+  --role-session-name reader-test
+# Use returned credentials to read from either region
 ```
 
 ## Cleanup
 
-```
+```bash
 cdk destroy
 ```
 
--- a/dynamodb-cross-account-replication-cdk/README.md
+++ b/dynamodb-cross-account-replication-cdk/README.md
@@ -1,6 +1,6 @@
 # Amazon DynamoDB Global Tables with Cross-Account Read Access
 
-This pattern deploys a DynamoDB Global Table (same-account, multi-region) and an IAM role that grants a separate AWS account read-only access to the table in both regions.
+This pattern deploys an Amazon DynamoDB Global Table (same-account, multi-region) and an IAM role that grants a separate AWS account read-only access to the table in both regions.
 
 Learn more about this pattern at Serverless Land Patterns: https://serverlessland.com/patterns/dynamodb-cross-account-replication-cdk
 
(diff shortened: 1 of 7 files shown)


NithinChandranR-AWS:
Hi @biswanathmukherjee 👋 This shows **DynamoDB Global Tables cross-account replication** — a unique integration pattern for multi-account architectures. No existing pattern covers cross-account DynamoDB replication. Deployed and tested.

dynamodb-cross-account-replication-cdk/lib/dynamodb-cross-account-replication-stack.ts
```diff
@@ -0,0 +1,52 @@
...
+    });
+
+    // IAM role for cross-account access to the replica
+    const crossAccountRole = new iam.Role(this, 'CrossAccountReplicaRole', {
+      assumedBy: new iam.AccountPrincipal(replicaAccountId.valueAsString),
+      description: 'Allows replica account to read from the global table replica',
+    });
+
+    crossAccountRole.addToPolicy(new iam.PolicyStatement({
+      actions: [
```
parikhudit: Cross-account read role cannot access the replica region
`table.tableArn` resolves to the **source-region** (us-east-1 by default) ARN. DynamoDB IAM resource ARNs are region-scoped `arn:aws:dynamodb:<region>:<account>:table/<name>`. When the replica account assumes this role and runs `GetItem` against the `us-west-2` replica, IAM evaluates against the `us-west-2` ARN, which is not in the policy → `AccessDenied`. The pattern's stated purpose (cross-account read from the replica) seems functionally broken.

dynamodb-cross-account-replication-cdk/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+1. Clone and navigate to the pattern:
+    ```
+    cd serverless-patterns/dynamodb-cross-account-replication-cdk
+    npm install
+    ```
+2. Deploy with the replica account ID:
+    ```
+    cdk deploy --parameters ReplicaAccountId=123456789012 --parameters ReplicaRegion=us-west-2
+    ```
```
parikhudit: `--parameters ReplicaRegion=us-west-2` is documented but the stack does not declare it.

`lib/dynamodb-cross-account-replication-stack.ts` only defines `ReplicaAccountId`. The replica region is the literal `'us-west-2'`. Users following this command will either get a CloudFormation error ("Parameter `ReplicaRegion` does not exist in the template") or have the parameter silently ignored. They think they're choosing a region but they aren't. You may want to add the parameter to the stack and use it

dynamodb-cross-account-replication-cdk/lib/dynamodb-cross-account-replication-stack.ts
```diff
@@ -0,0 +1,52 @@
...
+      sortKey: { name: 'SK', type: dynamodb.AttributeType.STRING },
+      billing: dynamodb.Billing.onDemand(),
+      pointInTimeRecovery: true,
+      removalPolicy: cdk.RemovalPolicy.DESTROY,
+      replicas: [
+        {
+          region: 'us-west-2',
+        },
+      ],
+    });
```
parikhudit, line 26: Title says "cross-account replication" but the implementation is same-account, multi-region.
`TableV2.replicas` only sets a region. There's no replica account property. The replica is created in the **same** AWS account as the source, in another region. True cross-account Global Tables require `TableV2MultiAccountReplica` in the replica account's stack, with `replicaSourceTable` referencing the source. 

Possible ways to resolve:

* **Reframe** as "cross-account READ over a same-account multi-region replica". Keep the current architecture, rename the pattern (`dynamodb-cross-account-read-cdk`?), and update title + README to make this explicit. The cross-account aspect is then only the IAM role.
* **Implement true multi-account Global Tables** by spliting into two stacks

dynamodb-cross-account-replication-cdk/lib/dynamodb-cross-account-replication-stack.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    // IAM role for cross-account access to the replica
+    const crossAccountRole = new iam.Role(this, 'CrossAccountReplicaRole', {
+      assumedBy: new iam.AccountPrincipal(replicaAccountId.valueAsString),
```
parikhudit, line 30: Trust policy allows any principal in the replica account; no `sts:ExternalId` or principal scoping.**

`iam.AccountPrincipal(replicaAccountId)` trusts the entire replica account. AWS recommends adding either an `sts:ExternalId` condition or a specific principal ARN to defend against confused-deputy patterns and to scope cross-account trust to a specific consumer.

dynamodb-cross-account-replication-cdk/lib/dynamodb-cross-account-replication-stack.ts
```diff
@@ -0,0 +1,52 @@
...
+export class DynamodbCrossAccountReplicationStack extends cdk.Stack {
+  constructor(scope: Construct, id: string, props?: cdk.StackProps) {
+    super(scope, id, props);
+
+    const replicaAccountId = new cdk.CfnParameter(this, 'ReplicaAccountId', {
+      type: 'String',
+      description: 'AWS Account ID for the replica table',
+    });
+
+    const table = new dynamodb.TableV2(this, 'SourceTable', {
```
parikhudit, line 18: KMS expectations for cross-account reads are not addressed. `TableV2` defaults to AWS-owned KMS keys, which work transparently across accounts. Real-world cross-account replication usually uses customer-managed KMS keys for compliance and at that point the replica account principal needs `kms:Decrypt` (and replication needs `kms:GenerateDataKey`/`kms:Encrypt`) on the key. Pattern docs should at minimum call this out so users don't get stuck on `KMSAccessDeniedException` after swapping in a CMK.

dynamodb-cross-account-replication-cdk/lib/dynamodb-cross-account-replication-stack.ts
NithinChandranR-AWS: You're right -- table.tableArn is source-region-scoped. Fixed by adding explicit replica region ARNs to the policy so the reader account can access the table in both regions.

dynamodb-cross-account-replication-cdk/README.md
NithinChandranR-AWS: Good catch. TableV2 doesn't accept tokens for replica regions (throws ReplicaTableRegionCannotBeToken), so CfnParameter won't work here. Switched to CDK context (-c replicaRegion=us-west-2) which resolves at synth time. README updated to match.

dynamodb-cross-account-replication-cdk/lib/dynamodb-cross-account-replication-stack.ts
NithinChandranR-AWS: True -- it trusts the entire account. For a sample pattern this keeps it simple. Adding ExternalId or role-name conditions would make it more production-ready but also more complex to demo. Happy to add an ExternalId condition if you feel strongly about it.

dynamodb-cross-account-replication-cdk/lib/dynamodb-cross-account-replication-stack.ts
NithinChandranR-AWS: The table uses the default AWS-owned key (no explicit KMS config), so cross-account reads work without any key policy changes. Added a note in the README that if customers switch to CMK encryption they'll need to grant kms:Decrypt to the cross-account role.

dynamodb-cross-account-replication-cdk/lib/dynamodb-cross-account-replication-stack.ts
NithinChandranR-AWS: Agreed -- reframed the whole pattern. Title is now "DynamoDB Global Tables with Cross-Account Read Access", README explicitly calls out that replication is same-account and the cross-account piece is the IAM role.

dynamodb-cross-account-replication-cdk/lib/dynamodb-cross-account-replication-stack.ts
```diff
@@ -0,0 +1,52 @@
...
+      description: 'Allows replica account to read from the global table replica',
+    });
+
+    crossAccountRole.addToPolicy(new iam.PolicyStatement({
+      actions: [
+        'dynamodb:GetItem',
+        'dynamodb:Query',
+        'dynamodb:Scan',
+        'dynamodb:BatchGetItem',
+      ],
```
parikhudit, line 39: `dynamodb:Scan` on a least-privilege read role. The role description is "Allows replica account to read from the global table replica." `Scan` reads every item broader than required for typical cross-account read use cases and a budget hazard. Drop unless explicitly needed.

dynamodb-cross-account-replication-cdk/lib/dynamodb-cross-account-replication-stack.ts
```diff
@@ -0,0 +1,52 @@
...
+        {
+          region: 'us-west-2',
+        },
+      ],
+    });
+
+    // IAM role for cross-account access to the replica
+    const crossAccountRole = new iam.Role(this, 'CrossAccountReplicaRole', {
+      assumedBy: new iam.AccountPrincipal(replicaAccountId.valueAsString),
+      description: 'Allows replica account to read from the global table replica',
```
parikhudit: `dynamodb:Scan` on a least-privilege read role. The role description is "Allows replica account to read from the global table replica." `Scan` reads every item broader than required for typical cross-account read use cases and a budget hazard. Drop unless explicitly needed.

dynamodb-cross-account-replication-cdk/README.md
```diff
@@ -0,0 +1,60 @@
...
+* [AWS CLI](https://docs.aws.amazon.com/cli/latest/userguide/install-cliv2.html) installed and configured
+* [AWS CDK](https://docs.aws.amazon.com/cdk/latest/guide/cli.html) installed
+* [Node.js](https://nodejs.org/en/download/) installed
+* Two AWS accounts (source and replica)
+
+## Deployment Instructions
+
+1. Clone and navigate to the pattern:
+    ```
+    cd serverless-patterns/dynamodb-cross-account-replication-cdk
```
parikhudit, line 49: `cdk bootstrap` is missing from Requirements. New users may hit AWS CDK bootstrap errors immediately.

NithinChandranR-AWS:
Based on your feedback, I've split this into two complementary patterns:

1. **This PR (#3099)** — reframed as "DynamoDB Global Tables with Cross-Account Read Access" (same-account multi-region replication + cross-account IAM role for reading)
2. **PR #3126** — true multi-account replication using `TableV2MultiAccountReplica` (two stacks, two accounts, automatic cross-account replication): https://github.com/aws-samples/serverless-patterns/pull/3126

Both deployed and tested end-to-end. Thanks for pushing me to get this right.

dynamodb-cross-account-replication-cdk/README.md
```diff
@@ -0,0 +1,60 @@
...
+    ```
+
+## How it works
+
+- A DynamoDB Global Table is created with a replica in the specified region
+- DynamoDB automatically replicates all writes to the replica with sub-second latency
+- A cross-account IAM role allows the replica account to assume and read from the table
+- Point-in-time recovery is enabled for data protection
+
+## Testing
```
parikhudit, line 57: No cross-account read test in the Testing section. The current Testing section only shows reads from the **same account**. The pattern's headline value is cross-account read, so please demonstrate it.

dynamodb-cross-account-replication-cdk/lib/dynamodb-cross-account-replication-stack.ts
NithinChandranR-AWS: Good point -- dropped Scan. GetItem, Query, and BatchGetItem cover the typical cross-account read use cases.

dynamodb-cross-account-replication-cdk/README.md
NithinChandranR-AWS: Added to requirements. Thanks.

dynamodb-cross-account-replication-cdk/README.md
```diff
@@ -0,0 +1,60 @@
...
+# Read from replica region (same account)
+aws dynamodb get-item --table-name <TableName> \
+  --key '{"PK":{"S":"user#123"},"SK":{"S":"profile"}}' \
+  --region us-west-2
+```
+
+## Cleanup
+
+```
+cdk destroy
```
parikhudit, line 91: Cleanup must warn that `cdk destroy` deletes the replica and its data. `RemovalPolicy.DESTROY` plus `cdk destroy` deletes the source table and all replicas with no recovery. Worth a clear heads-up so testers don't lose data unwillingly. Do confirm this though.

dynamodb-cross-account-replication-cdk/lib/dynamodb-cross-account-replication-stack.ts
NithinChandranR-AWS: Same as above -- Scan dropped.

dynamodb-cross-account-replication-cdk/README.md
NithinChandranR-AWS: Fair -- added a cross-account read example using sts assume-role in the Testing section.

dynamodb-cross-account-replication-cdk/README.md
NithinChandranR-AWS: Good call -- added a warning note in the Cleanup section about data deletion with RemovalPolicy.DESTROY.

dynamodb-cross-account-replication-cdk/bin/app.ts
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env node
+import 'source-map-support/register';
+import * as cdk from 'aws-cdk-lib';
+import { DynamodbCrossAccountReplicationStack } from '../lib/dynamodb-cross-account-replication-stack';
+
+const app = new cdk.App();
+new DynamodbCrossAccountReplicationStack(app, 'DynamodbCrossAccountReplicationStack', {
+  env: {
+    account: process.env.CDK_DEFAULT_ACCOUNT || '123456789012',
```
parikhudit: Hardcoded fallback account `'123456789012'` is misleading. Synthesizing against the docs-example account when `CDK_DEFAULT_ACCOUNT` is unset will produce templates that look fine in `cdk synth` but fail at deploy in confusing ways. Either fall back to environment-agnostic synthesis or fail loudly:

dynamodb-cross-account-replication-cdk/lib/dynamodb-cross-account-replication-stack.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    const table = new dynamodb.TableV2(this, 'SourceTable', {
+      partitionKey: { name: 'PK', type: dynamodb.AttributeType.STRING },
+      sortKey: { name: 'SK', type: dynamodb.AttributeType.STRING },
+      billing: dynamodb.Billing.onDemand(),
+      pointInTimeRecovery: true,
+      removalPolicy: cdk.RemovalPolicy.DESTROY,
+      replicas: [
+        {
+          region: 'us-west-2',
```
parikhudit: Replica region `'us-west-2'` is hardcoded in two places

dynamodb-cross-account-replication-cdk/README.md
```diff
@@ -0,0 +1,60 @@
...
+    npm install
+    ```
+2. Deploy with the replica account ID:
+    ```
+    cdk deploy --parameters ReplicaAccountId=123456789012 --parameters ReplicaRegion=us-west-2
+    ```
+
+## How it works
+
+- A DynamoDB Global Table is created with a replica in the specified region
```
parikhudit: Use "Amazon DynamoDB" on first body reference. The title correctly uses "Amazon DynamoDB"; the first body paragraph and the "How it works" bullet immediately drop to "DynamoDB Global Table". Per AWS naming guidance the first body reference should re-introduce the full service name.

dynamodb-cross-account-replication-cdk/lib/dynamodb-cross-account-replication-stack.ts
```diff
@@ -0,0 +1,52 @@
...
+      type: 'String',
+      description: 'AWS Account ID for the replica table',
+    });
+
+    const table = new dynamodb.TableV2(this, 'SourceTable', {
+      partitionKey: { name: 'PK', type: dynamodb.AttributeType.STRING },
+      sortKey: { name: 'SK', type: dynamodb.AttributeType.STRING },
+      billing: dynamodb.Billing.onDemand(),
+      pointInTimeRecovery: true,
+      removalPolicy: cdk.RemovalPolicy.DESTROY,
```
parikhudit, line 24: The pattern frames itself as a cross-account replication solution and turns on PITR, both signal a production lean. Pairing that with `DESTROY` is a foot-gun. Either flip the default to `RETAIN`, or keep `DESTROY` for testability and call it out clearly in source + README.

dynamodb-cross-account-replication-cdk/lib/dynamodb-cross-account-replication-stack.ts
```diff
@@ -0,0 +1,52 @@
...
+    const replicaAccountId = new cdk.CfnParameter(this, 'ReplicaAccountId', {
+      type: 'String',
+      description: 'AWS Account ID for the replica table',
+    });
+
+    const table = new dynamodb.TableV2(this, 'SourceTable', {
+      partitionKey: { name: 'PK', type: dynamodb.AttributeType.STRING },
+      sortKey: { name: 'SK', type: dynamodb.AttributeType.STRING },
+      billing: dynamodb.Billing.onDemand(),
+      pointInTimeRecovery: true,
```
parikhudit: You may want to refer and use `pointInTimeRecoverySpecification` instead, please check.

dynamodb-cross-account-replication-cdk/bin/app.ts
NithinChandranR-AWS: Fixed -- app.ts now throws with a clear error message if CDK_DEFAULT_ACCOUNT is unset. No more silent fallback to a placeholder.

dynamodb-cross-account-replication-cdk/lib/dynamodb-cross-account-replication-stack.ts
NithinChandranR-AWS: It's a CDK context value with us-west-2 as fallback (not a string literal baked into logic). Users override via -c replicaRegion=eu-west-1. The fallback is intentional so the pattern works out-of-the-box for quick testing.

dynamodb-cross-account-replication-cdk/README.md
NithinChandranR-AWS: Fixed -- first body reference now uses "Amazon DynamoDB Global Table".

dynamodb-cross-account-replication-cdk/lib/dynamodb-cross-account-replication-stack.ts
NithinChandranR-AWS: Fair point. Added inline comment in code explaining DESTROY is for sample pattern testability. Cleanup section already warns about data deletion. Kept DESTROY since RETAIN would leave orphaned tables for users just trying the pattern.

dynamodb-cross-account-replication-cdk/lib/dynamodb-cross-account-replication-stack.ts
NithinChandranR-AWS: Already using pointInTimeRecoverySpecification in the latest commit -- the deprecated pointInTimeRecovery was from an earlier version.

NithinChandranR-AWS:

NithinChandranR-AWS:

NithinChandranR-AWS:

NithinChandranR-AWS:

PR #3100, opened 2026-05-06
New pattern - eventbridge-cloudtrail-dataplane-cdk

NithinChandranR-AWS:
## Description
First pattern for **Amazon EventBridge data plane logging to AWS CloudTrail** (launched May 5, 2026).

## What it does
- Enables CloudTrail data plane event logging for EventBridge
- EventBridge rule captures PutEvents API calls from CloudTrail
- Lambda function alerts on caller identity, source IP, event bus, and entry count

## Testing
Deployed and tested. CloudTrail trail created, EventBridge rule configured, Lambda processor with JSON logging verified.

---

### DIFF
--- a/eventbridge-cloudtrail-dataplane-cdk/README.md
+++ b/eventbridge-cloudtrail-dataplane-cdk/README.md
@@ -1,0 +1,58 @@
+# Amazon EventBridge Data Plane Logging with AWS CloudTrail
+
+This pattern enables CloudTrail data plane logging for Amazon EventBridge and triggers a Lambda function when PutEvents API calls are detected, providing security and operational visibility into event bus activity.
+
+Learn more about this pattern at Serverless Land Patterns: https://serverlessland.com/patterns/eventbridge-cloudtrail-dataplane-cdk
+
+Important: this application uses various AWS services and there are costs associated with these services after the Free Tier usage - please see the [AWS Pricing page](https://aws.amazon.com/pricing/) for details.
+
+## Requirements
+
+* [AWS CLI](https://docs.aws.amazon.com/cli/latest/userguide/install-cliv2.html) installed and configured
+* [AWS CDK](https://docs.aws.amazon.com/cdk/latest/guide/cli.html) installed
+* [Node.js](https://nodejs.org/en/download/) installed
+
+## Deployment Instructions
+
+1. Clone and navigate to the pattern:
+    ```
+    cd serverless-patterns/eventbridge-cloudtrail-dataplane-cdk
+    npm install
+    ```
+2. Deploy:
+    ```
+    cdk deploy
+    ```
+
+## How it works
+
+- A CloudTrail trail is created with data event logging enabled
+- EventBridge data plane API calls (PutEvents) are now logged to CloudTrail (new May 2026 feature)
+- An EventBridge rule captures these CloudTrail events matching `aws.events` source with `PutEvents` event name
+- A Lambda function processes the events, logging the caller identity, source IP, event bus, and entry count
+- This enables security teams to audit who is putting events to which bus
+
+## Testing
+
+```bash
+# Put a test event to the default event bus
+aws events put-events --entries '[{"Source":"test.app","DetailType":"TestEvent","Detail":"{\"key\":\"value\"}"}]'
+
+# Check Lambda logs (allow ~5 minutes for CloudTrail delivery)
+aws logs tail /aws/lambda/$(aws cloudformation describe-stacks \
+  --stack-name EventbridgeCloudtrailDataplaneStack \
+  --query 'Stacks[0].Outputs[?OutputKey==`ProcessorFunctionName`].OutputValue' --output text) \
+  --follow
+```
+
+## Cleanup
+
+```
+cdk destroy
+```
+
+---
+
+Copyright 2026 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+SPDX-License-Identifier: MIT-0
--- a/eventbridge-cloudtrail-dataplane-cdk/README.md
+++ b/eventbridge-cloudtrail-dataplane-cdk/README.md
@@ -6,11 +6,13 @@ Learn more about this pattern at Serverless Land Patterns: https://serverlesslan
 
 Important: this application uses various AWS services and there are costs associated with these services after the Free Tier usage - please see the [AWS Pricing page](https://aws.amazon.com/pricing/) for details.
 
+> **Note:** CloudTrail data events are billed separately from management events. See [AWS CloudTrail Pricing](https://aws.amazon.com/cloudtrail/pricing/) for current data event rates.
+
 ## Requirements
 
 * [AWS CLI](https://docs.aws.amazon.com/cli/latest/userguide/install-cliv2.html) installed and configured
 * [AWS CDK](https://docs.aws.amazon.com/cdk/latest/guide/cli.html) installed
-* [Node.js](https://nodejs.org/en/download/) installed
+* [Node.js 20+](https://nodejs.org/en/download/) installed
 
 ## Deployment Instructions
 
@@ -19,14 +21,18 @@ Important: this application uses various AWS services and there are costs associ
     cd serverless-patterns/eventbridge-cloudtrail-dataplane-cdk
     npm install
     ```
-2. Deploy:
+2. Bootstrap CDK (first-time CDK users only):
+    ```
+    cdk bootstrap
+    ```
+3. Deploy:
     ```
     cdk deploy
     ```
 
 ## How it works
 
-- A CloudTrail trail is created with data event logging enabled
+- A CloudTrail trail is created with data event logging enabled for EventBridge event buses (`AWS::Events::EventBus`)
 - EventBridge data plane API calls (PutEvents) are now logged to CloudTrail (new May 2026 feature)
 - An EventBridge rule captures these CloudTrail events matching `aws.events` source with `PutEvents` event name
 - A Lambda function processes the events, logging the caller identity, source IP, event bus, and entry count
--- a/eventbridge-cloudtrail-dataplane-cdk/lib/eventbridge-cloudtrail-dataplane-stack.ts
+++ b/eventbridge-cloudtrail-dataplane-cdk/lib/eventbridge-cloudtrail-dataplane-stack.ts
@@ -1,0 +1,58 @@
+import * as cdk from 'aws-cdk-lib';
+import * as events from 'aws-cdk-lib/aws-events';
+import * as targets from 'aws-cdk-lib/aws-events-targets';
+import * as lambda from 'aws-cdk-lib/aws-lambda';
+import * as logs from 'aws-cdk-lib/aws-logs';
+import * as cloudtrail from 'aws-cdk-lib/aws-cloudtrail';
+import * as s3 from 'aws-cdk-lib/aws-s3';
+import { Construct } from 'constructs';
+
+export class EventbridgeCloudtrailDataplaneStack extends cdk.Stack {
+  constructor(scope: Construct, id: string, props?: cdk.StackProps) {
+    super(scope, id, props);
+
+    // S3 bucket for CloudTrail logs
+    const trailBucket = new s3.Bucket(this, 'TrailBucket', {
+      removalPolicy: cdk.RemovalPolicy.DESTROY,
+      autoDeleteObjects: true,
+      enforceSSL: true,
+    });
+
+    // CloudTrail trail with data events for EventBridge
+    const trail = new cloudtrail.Trail(this, 'EventBridgeDataPlaneTrail', {
+      bucket: trailBucket,
+      trailName: 'eventbridge-dataplane-trail',
+      isMultiRegionTrail: false,
+    });
+
+    // Enable EventBridge data plane events logging
+    trail.addEventSelector(cloudtrail.DataResourceType.LAMBDA_FUNCTION, ['arn:aws:lambda']);
+
+    // Lambda function to process CloudTrail events
+    const processor = new lambda.Function(this, 'EventProcessor', {
+      runtime: lambda.Runtime.NODEJS_20_X,
+      handler: 'index.handler',
+      code: lambda.Code.fromAsset('src'),
+      timeout: cdk.Duration.seconds(10),
+      loggingFormat: lambda.LoggingFormat.JSON,
+    });
+
+    // EventBridge rule to capture EventBridge PutEvents API calls from CloudTrail
+    const rule = new events.Rule(this, 'DataPlaneRule', {
+      eventPattern: {
+        source: ['aws.events'],
+        detailType: ['AWS API Call via CloudTrail'],
+        detail: {
+          eventSource: ['events.amazonaws.com'],
+          eventName: ['PutEvents'],
+        },
+      },
+    });
+
+    rule.addTarget(new targets.LambdaFunction(processor));
+
+    new cdk.CfnOutput(this, 'ProcessorFunctionName', { value: processor.functionName });
+    new cdk.CfnOutput(this, 'TrailBucketName', { value: trailBucket.bucketName });
+    new cdk.CfnOutput(this, 'RuleName', { value: rule.ruleName });
+  }
+}
--- a/eventbridge-cloudtrail-dataplane-cdk/lib/eventbridge-cloudtrail-dataplane-stack.ts
+++ b/eventbridge-cloudtrail-dataplane-cdk/lib/eventbridge-cloudtrail-dataplane-stack.ts
@@ -2,31 +2,61 @@ import * as cdk from 'aws-cdk-lib';
 import * as events from 'aws-cdk-lib/aws-events';
 import * as targets from 'aws-cdk-lib/aws-events-targets';
 import * as lambda from 'aws-cdk-lib/aws-lambda';
+import * as destinations from 'aws-cdk-lib/aws-lambda-destinations';
 import * as logs from 'aws-cdk-lib/aws-logs';
 import * as cloudtrail from 'aws-cdk-lib/aws-cloudtrail';
 import * as s3 from 'aws-cdk-lib/aws-s3';
+import * as sqs from 'aws-cdk-lib/aws-sqs';
 import { Construct } from 'constructs';
 
 export class EventbridgeCloudtrailDataplaneStack extends cdk.Stack {
   constructor(scope: Construct, id: string, props?: cdk.StackProps) {
     super(scope, id, props);
 
+    // S3 access logs bucket
+    const accessLogsBucket = new s3.Bucket(this, 'AccessLogsBucket', {
+      removalPolicy: cdk.RemovalPolicy.DESTROY,
+      autoDeleteObjects: true,
+      enforceSSL: true,
+      encryption: s3.BucketEncryption.S3_MANAGED,
+      blockPublicAccess: s3.BlockPublicAccess.BLOCK_ALL,
+    });
+
     // S3 bucket for CloudTrail logs
     const trailBucket = new s3.Bucket(this, 'TrailBucket', {
       removalPolicy: cdk.RemovalPolicy.DESTROY,
       autoDeleteObjects: true,
       enforceSSL: true,
+      encryption: s3.BucketEncryption.S3_MANAGED,
+      blockPublicAccess: s3.BlockPublicAccess.BLOCK_ALL,
+      serverAccessLogsBucket: accessLogsBucket,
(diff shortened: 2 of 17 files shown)


NithinChandranR-AWS:
Hi @biswanathmukherjee 👋 This demonstrates **CloudTrail data plane events → EventBridge** — a unique integration using the new data event filtering (2026). First pattern showing S3 data-plane events triggering Lambda via EventBridge. Deployed and tested.

eventbridge-cloudtrail-dataplane-cdk/lib/eventbridge-cloudtrail-dataplane-stack.ts
```diff
@@ -0,0 +1,58 @@
...
+    const trail = new cloudtrail.Trail(this, 'EventBridgeDataPlaneTrail', {
+      bucket: trailBucket,
+      trailName: 'eventbridge-dataplane-trail',
+      isMultiRegionTrail: false,
+    });
+
+    // Enable EventBridge data plane events logging
+    trail.addEventSelector(cloudtrail.DataResourceType.LAMBDA_FUNCTION, ['arn:aws:lambda']);
+
+    // Lambda function to process CloudTrail events
```
parikhudit, line 61: Wrong CloudTrail data event resource type; this trail does not capture EventBridge `PutEvents`. `addEventSelector(DataResourceType.LAMBDA_FUNCTION, ['arn:aws:lambda'])` adds a *basic* event selector that logs **Lambda Invoke** data events for every Lambda function in the account/region. EventBridge `PutEvents` data events require an **advanced** event selector with `resources.type = AWS::Events::EventBus`, per the [EventBridge / CloudTrail integration docs](https://docs.aws.amazon.com/eventbridge/latest/userguide/logging-using-cloudtrail.html#cloudtrail-event-examples). The CDK L2 `DataResourceType` enum only exposes `LAMBDA_FUNCTION` and `S3_OBJECT`, so we need `CfnTrail` with `AdvancedEventSelectors`. Kindly check and confirm.

parikhudit: I'd also suggest a quick smoke-test command in the README that asserts the Lambda actually logged a record.

eventbridge-cloudtrail-dataplane-cdk/lib/eventbridge-cloudtrail-dataplane-stack.ts
```diff
@@ -0,0 +1,58 @@
...
+export class EventbridgeCloudtrailDataplaneStack extends cdk.Stack {
+  constructor(scope: Construct, id: string, props?: cdk.StackProps) {
+    super(scope, id, props);
+
+    // S3 bucket for CloudTrail logs
+    const trailBucket = new s3.Bucket(this, 'TrailBucket', {
+      removalPolicy: cdk.RemovalPolicy.DESTROY,
+      autoDeleteObjects: true,
+      enforceSSL: true,
+    });
```
parikhudit, line 34: Set explicit `encryption` and `blockPublicAccess` on the trail bucket. This bucket holds CloudTrail audit logs, so best to be explicit even though Amazon S3 now defaults to SSE-S3 at the API.

eventbridge-cloudtrail-dataplane-cdk/lib/eventbridge-cloudtrail-dataplane-stack.ts
```diff
@@ -0,0 +1,58 @@
...
+        source: ['aws.events'],
+        detailType: ['AWS API Call via CloudTrail'],
+        detail: {
+          eventSource: ['events.amazonaws.com'],
+          eventName: ['PutEvents'],
+        },
+      },
+    });
+
+    rule.addTarget(new targets.LambdaFunction(processor));
```
parikhudit: Add a DLQ to the EventBridge target. Without `deadLetterQueue` on the `LambdaFunction` target, failed invocations would end up in the void after EventBridge's default retries.

eventbridge-cloudtrail-dataplane-cdk/lib/eventbridge-cloudtrail-dataplane-stack.ts
```diff
@@ -0,0 +1,58 @@
...
+    trail.addEventSelector(cloudtrail.DataResourceType.LAMBDA_FUNCTION, ['arn:aws:lambda']);
+
+    // Lambda function to process CloudTrail events
+    const processor = new lambda.Function(this, 'EventProcessor', {
+      runtime: lambda.Runtime.NODEJS_20_X,
+      handler: 'index.handler',
+      code: lambda.Code.fromAsset('src'),
+      timeout: cdk.Duration.seconds(10),
+      loggingFormat: lambda.LoggingFormat.JSON,
+    });
```
parikhudit, line 71: Configure `onFailure` and `retryAttempts` on the async Lambda. Even with the EventBridge-side DLQ, function-level destinations give clearer signal for code-level failures.

eventbridge-cloudtrail-dataplane-cdk/lib/eventbridge-cloudtrail-dataplane-stack.ts
```diff
@@ -0,0 +1,58 @@
...
+      removalPolicy: cdk.RemovalPolicy.DESTROY,
+      autoDeleteObjects: true,
+      enforceSSL: true,
+    });
+
+    // CloudTrail trail with data events for EventBridge
+    const trail = new cloudtrail.Trail(this, 'EventBridgeDataPlaneTrail', {
+      bucket: trailBucket,
+      trailName: 'eventbridge-dataplane-trail',
+      isMultiRegionTrail: false,
```
parikhudit: `isMultiRegionTrail: false` misses cross-region `PutEvents`. For an audit/security pattern, the trail should capture all regions of the account. Console-created trails are multi-region by default. Suggest `isMultiRegionTrail: true`. Also you may want to call out in README that EventBridge rules are region-local, multi-region detection requires deploying the rule in each region or fanning in via cross-region buses.

eventbridge-cloudtrail-dataplane-cdk/README.md
```diff
@@ -0,0 +1,58 @@
...
+* [AWS CDK](https://docs.aws.amazon.com/cdk/latest/guide/cli.html) installed
+* [Node.js](https://nodejs.org/en/download/) installed
+
+## Deployment Instructions
+
+1. Clone and navigate to the pattern:
+    ```
+    cd serverless-patterns/eventbridge-cloudtrail-dataplane-cdk
+    npm install
+    ```
```
parikhudit, line 23: Add `cdk bootstrap` for first time users

eventbridge-cloudtrail-dataplane-cdk/lib/eventbridge-cloudtrail-dataplane-stack.ts
```diff
@@ -0,0 +1,58 @@
...
+    const trailBucket = new s3.Bucket(this, 'TrailBucket', {
+      removalPolicy: cdk.RemovalPolicy.DESTROY,
+      autoDeleteObjects: true,
+      enforceSSL: true,
+    });
+
+    // CloudTrail trail with data events for EventBridge
+    const trail = new cloudtrail.Trail(this, 'EventBridgeDataPlaneTrail', {
+      bucket: trailBucket,
+      trailName: 'eventbridge-dataplane-trail',
```
parikhudit: Please drop the hardcoded `trailName`. 
Although `trailName: 'eventbridge-dataplane-trail'` is unique per account/region. A second deployment of the stack (e.g. to test a change in a different stack name) could fail with `Trail already exists`. For a sample, the simplest fix is to remove the property and let CDK auto-generate a unique name. If a stable name is desired, expose it as a `CfnParameter`.

eventbridge-cloudtrail-dataplane-cdk/lib/eventbridge-cloudtrail-dataplane-stack.ts
```diff
@@ -0,0 +1,58 @@
...
+
+export class EventbridgeCloudtrailDataplaneStack extends cdk.Stack {
+  constructor(scope: Construct, id: string, props?: cdk.StackProps) {
+    super(scope, id, props);
+
+    // S3 bucket for CloudTrail logs
+    const trailBucket = new s3.Bucket(this, 'TrailBucket', {
+      removalPolicy: cdk.RemovalPolicy.DESTROY,
+      autoDeleteObjects: true,
+      enforceSSL: true,
```
parikhudit, line 29: The bucket holding CloudTrail logs has no serverAccessLogsBucket configured. For an audit-trail bucket, S3 server access logging adds a second-tier audit record of who accessed/modified the audit logs themselves. This is a defense-in-depth recommendation rather than a hard requirement.

eventbridge-cloudtrail-dataplane-cdk/README.md
```diff
@@ -0,0 +1,58 @@
+# Amazon EventBridge Data Plane Logging with AWS CloudTrail
+
+This pattern enables CloudTrail data plane logging for Amazon EventBridge and triggers a Lambda function when PutEvents API calls are detected, providing security and operational visibility into event bus activity.
+
+Learn more about this pattern at Serverless Land Patterns: https://serverlessland.com/patterns/eventbridge-cloudtrail-dataplane-cdk
+
+Important: this application uses various AWS services and there are costs associated with these services after the Free Tier usage - please see the [AWS Pricing page](https://aws.amazon.com/pricing/) for details.
```
parikhudit, line 7: Recommendation: Mention that CloudTrail data events are billed separately and link [pricing](https://aws.amazon.com/cloudtrail/pricing)

eventbridge-cloudtrail-dataplane-cdk/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Requirements
+
+* [AWS CLI](https://docs.aws.amazon.com/cli/latest/userguide/install-cliv2.html) installed and configured
+* [AWS CDK](https://docs.aws.amazon.com/cdk/latest/guide/cli.html) installed
+* [Node.js](https://nodejs.org/en/download/) installed
```
parikhudit: Specify Node.js 18+ or appropriate in Requirements (function uses Node.js 20.x)

eventbridge-cloudtrail-dataplane-cdk/src/index.js
```diff
@@ -0,0 +1,15 @@
+exports.handler = async (event) => {
+  const detail = event.detail || {};
+  console.log(JSON.stringify({
```
parikhudit: using`console.log(JSON.stringify(...))` here, on Node.js 20 with `LoggingFormat.JSON` you can use `context.logger.info({...})` directly. The `return { statusCode: 200 }` would be unused for async EventBridge invocations

eventbridge-cloudtrail-dataplane-cdk/lib/eventbridge-cloudtrail-dataplane-stack.ts
```diff
@@ -0,0 +1,58 @@
...
+    // Enable EventBridge data plane events logging
+    trail.addEventSelector(cloudtrail.DataResourceType.LAMBDA_FUNCTION, ['arn:aws:lambda']);
+
+    // Lambda function to process CloudTrail events
+    const processor = new lambda.Function(this, 'EventProcessor', {
+      runtime: lambda.Runtime.NODEJS_20_X,
+      handler: 'index.handler',
+      code: lambda.Code.fromAsset('src'),
+      timeout: cdk.Duration.seconds(10),
+      loggingFormat: lambda.LoggingFormat.JSON,
```
parikhudit, line 67: The processor function has no logRetention (or pre-created logGroup with retention) configured. The auto-created log group /aws/lambda/<fn> will retain logs forever, accruing CloudWatch Logs storage cost. So either set retention or give a callout in README and wherever applicable.

eventbridge-cloudtrail-dataplane-cdk/lib/eventbridge-cloudtrail-dataplane-stack.ts
parikhudit: After fixing above, you may want to drop or update the "allow ~5 minutes" note in the README test step, as CloudTrail data events would not typically take that long to reach EventBridge.

parikhudit:
Thanks for the submission. Heads-up that the repo already has s3-eventbridge ("S3 -> CloudTrail -> EventBridge") which is architecturally identical to this pattern, self-provisioned trail + data-event selector + EventBridge rule on AWS API Call via CloudTrail. There are few other similar patterns as well. The only thing that makes yours distinct is monitoring EventBridge's own PutEvents data plane.

Two issues to resolve before we can take it:

* Correctness (blocking): the trail uses a basic LAMBDA_FUNCTION selector, so it logs Lambda invokes, not EventBridge PutEvents. EventBridge data events need an advanced event selector with resources.type = AWS::Events::EventBus. As written, the rule may never fire. (Already added as comment)

* Differentiation: even once fixed, the pattern is very close to s3-eventbridge. Please add an EventBridge-specific angle that the other patterns do not cover e.g., auditing which producers publish to which buses, scoping the selector to specific named/custom buses, or a Lambda that validates/flags unexpected event sources, etc and update the README to explain why so. However, you would need to do research and find a unique pattern.

For these reasons, my inclination is towards enriching one of the existing patterns (as we have many), rather than adding a new one. Would love to hear your thoughts.

eventbridge-cloudtrail-dataplane-cdk/lib/eventbridge-cloudtrail-dataplane-stack.ts
NithinChandranR-AWS: Good catch — you're right, LAMBDA_FUNCTION doesn't capture EventBridge PutEvents at all. Switched to `AWS::Events::EventBus` via advanced event selectors (the CDK L2 `addEventSelector` doesn't support this resource type yet, so I used the CfnTrail escape hatch).

eventbridge-cloudtrail-dataplane-cdk/lib/eventbridge-cloudtrail-dataplane-stack.ts
NithinChandranR-AWS: Done — added `encryption: S3_MANAGED` and `blockPublicAccess: BLOCK_ALL` to both buckets.

eventbridge-cloudtrail-dataplane-cdk/lib/eventbridge-cloudtrail-dataplane-stack.ts
NithinChandranR-AWS: Added a shared SQS DLQ for the EventBridge target. Failed deliveries go there with 14-day retention.

eventbridge-cloudtrail-dataplane-cdk/lib/eventbridge-cloudtrail-dataplane-stack.ts
NithinChandranR-AWS: Done — set `retryAttempts: 2` and `onFailure: new SqsDestination(dlq)` on the Lambda.

eventbridge-cloudtrail-dataplane-cdk/lib/eventbridge-cloudtrail-dataplane-stack.ts
NithinChandranR-AWS: Fair point, flipped to `isMultiRegionTrail: true`.

eventbridge-cloudtrail-dataplane-cdk/README.md
NithinChandranR-AWS: Added a `cdk bootstrap` step in the deployment instructions.

eventbridge-cloudtrail-dataplane-cdk/lib/eventbridge-cloudtrail-dataplane-stack.ts
NithinChandranR-AWS: Removed — letting CloudFormation generate the name now.

eventbridge-cloudtrail-dataplane-cdk/lib/eventbridge-cloudtrail-dataplane-stack.ts
NithinChandranR-AWS: Added a dedicated `accessLogsBucket` with `serverAccessLogsBucket` pointing to it.

eventbridge-cloudtrail-dataplane-cdk/README.md
NithinChandranR-AWS: Added a callout box in the README noting CloudTrail data events are billed separately, with a link to the pricing page.

eventbridge-cloudtrail-dataplane-cdk/README.md
NithinChandranR-AWS: Updated to Node.js 20+.

eventbridge-cloudtrail-dataplane-cdk/src/index.js
NithinChandranR-AWS: Switched to `console.info` — with `loggingFormat: JSON` on the Lambda, CloudWatch will pick up the structured fields automatically.

eventbridge-cloudtrail-dataplane-cdk/lib/eventbridge-cloudtrail-dataplane-stack.ts
NithinChandranR-AWS: Added `logRetention: ONE_WEEK`. Keeps costs down for a demo pattern.

NithinChandranR-AWS:
Fair point — the overlap with s3-eventbridge is clear if you squint at the plumbing. The unique angle here is auditing EventBridge's own data-plane activity (which producers hit which buses, unexpected event sources), not S3 object events flowing through CloudTrail.

I've already fixed the blocking issue (switched to advanced event selectors with `AWS::Events::EventBus` resource type in the earlier commit). Let me reframe the README to lead with the "EventBridge bus audit" use case and drop the generic CloudTrail framing. That should make the differentiation obvious at a glance.

If after the rewrite you still feel it's too close, happy to close this one — no hard feelings.

eventbridge-cloudtrail-dataplane-cdk/lib/eventbridge-cloudtrail-dataplane-stack.ts
NithinChandranR-AWS: Good idea — I'll add a quick aws events put-events + aws logs tail assertion to the Testing section so folks can verify end-to-end without guessing.

NithinChandranR-AWS: You're right — data events via advanced selectors hit EventBridge much faster than the old management-event path. I'll update to ~30 seconds and add the smoke-test command so it's verifiable.

NithinChandranR-AWS:
Fair points on both. You're right that the basic event selector doesn't capture EventBridge PutEvents -- it needs an advanced selector with resources.type = AWS::Events::EventBus. I've already fixed that in the inline comments.

On differentiation -- I agree this is thin vs the existing s3-eventbridge patterns. Given your suggestion, I think the right move here is to either close this and enrich the existing pattern, or pivot to a genuinely distinct angle (e.g., auditing cross-account event sources on custom buses with a validation Lambda). Let me research what would add real value vs just being another trail+rule combo. I'm leaning toward closing this one honestly -- better to focus energy on patterns that fill a clear gap. What do you think?

eventbridge-cloudtrail-dataplane-cdk/lib/eventbridge-cloudtrail-dataplane-stack.ts
NithinChandranR-AWS: Good idea -- added a CLI verification command that checks the CloudTrail log group for recent events.

NithinChandranR-AWS: Updated the testing section to reference the CLI check instead of the wait-and-hope approach.

NithinChandranR-AWS:

NithinChandranR-AWS:

NithinChandranR-AWS:

NithinChandranR-AWS:

PR #3067, opened 2026-04-20
New pattern - bedrock-guardrails-cross-account-cdk

NithinChandranR-AWS:
## Bedrock Guardrails Account-Level Enforcement

Creates a Bedrock Guardrail, versions it, and enables account-level enforcement. The guardrail automatically applies to ALL Bedrock calls in the account without specifying guardrailIdentifier.

**Architecture:** CDK creates Guardrail → versions via AwsCustomResource → enforces account-wide → Test Lambda demonstrates automatic enforcement

**Key decisions:**
- AwsCustomResource for PutEnforcedGuardrailConfiguration (not in Lambda runtime SDK)
- Content filters (HATE, INSULTS, SEXUAL, VIOLENCE, MISCONDUCT, PROMPT_ATTACK) + denied topic (investment advice)
- Automatic cleanup on stack deletion

**Test results:**
- Safe prompt (What is Amazon S3?) → full response
- Violating prompt (What stocks should I buy?) → blocked with guardrail_intervened

TypeScript / AWS CDK

PR #3113, opened 2026-05-16
New pattern - apigw-lambda-bedrock-guardrails-cdk

NithinChandranR-AWS:
## Description
Deploy a REST API that invokes Amazon Bedrock with **application-level Guardrails** — per-request content and topic filtering.

## Application-level vs Account-level
This pattern applies guardrails per-request by passing `guardrailIdentifier` and `guardrailVersion` in each InvokeModel call. Different APIs can use different guardrails or skip them entirely. For account-wide enforcement, see `bedrock-guardrails-cross-account-cdk`.

## What it does
- API Gateway REST API → Lambda → Bedrock InvokeModel with Guardrails
- Content filters: sexual, violence, hate, insults, misconduct
- Topic filter: financial advice (DENY)
- Blocked requests return custom message without consuming model tokens

## Testing
Deployed and tested:
- ✅ Safe prompt ('What is Amazon S3?') → full response, guardrailAction: NONE
- ✅ Blocked prompt ('What stocks should I buy?') → blocked message, no tokens consumed

---

### DIFF
--- a/apigw-lambda-bedrock-guardrails-cdk/README.md
+++ b/apigw-lambda-bedrock-guardrails-cdk/README.md
@@ -1,0 +1,62 @@
+# API Gateway to Lambda to Amazon Bedrock with Application-Level Guardrails
+
+This pattern deploys a REST API that invokes Amazon Bedrock with **per-request Guardrails** — content and topic filtering applied at the application level on each individual API call.
+
+> **Application-level vs Account-level:** This pattern applies guardrails per-request by passing `guardrailIdentifier` and `guardrailVersion` in each InvokeModel call. This gives you fine-grained control — different APIs can use different guardrails, or skip them entirely. For account-wide enforcement that applies to ALL Bedrock calls automatically, see the [bedrock-guardrails-cross-account-cdk](../bedrock-guardrails-cross-account-cdk) pattern.
+
+Learn more about this pattern at Serverless Land Patterns: https://serverlessland.com/patterns/apigw-lambda-bedrock-guardrails-cdk
+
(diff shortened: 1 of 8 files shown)


bfreiberg:
Seems to be a duplicate of #3067

NithinChandranR-AWS:
Not a duplicate -- they're different layers of the same feature. #3067 is account-level enforcement (PutEnforcedGuardrailConfiguration -- guardrail applies to ALL Bedrock calls in the account without any code changes). This PR (#3113) is application-level guardrails (individual Lambda passes guardrailIdentifier in each Converse API call). Different use cases: #3067 for org-wide compliance, this one for app-specific content filtering where you choose which calls get guarded.

bfreiberg:
Thanks @NithinChandranR-AWS for the additional context. You're right that this isn't a duplicate of #3067 
Upon closer inspection, this submission unfortunately doesn't meet the bar for an architectural pattern. The serverless-patterns repository is intended to showcase reusable architectural patterns: distinct compositions of services, integrations, or invocation models that solve a non-trivial design problem. This submission configures Bedrock Guardrails by passing the `guardrailIdentifier` and `guardrailVersion` parameters on an `InvokeModel` (or `Converse`) call, which is a standard SDK feature rather than an architectural construct. There's no novel service-to-service wiring, event flow, or topology here that another builder couldn't get from the Bedrock API reference.

NithinChandranR-AWS:
Fair point -- you're right that the guardrail invocation itself is just an SDK parameter, not a novel architectural composition. I'll close this one. Thanks for the thorough explanation.

PR #3126, opened 2026-06-05
New serverless pattern - dynamodb-global-tables-multi-account-cdk

NithinChandranR-AWS:
### Description

Deploys a DynamoDB Global Table that replicates data across two separate AWS accounts using `TableV2MultiAccountReplica` (CDK 2.252.0+). Writes in Account A are automatically replicated to Account B with sub-second latency.

### Architecture

```
Account A (us-east-1)              Account B (us-west-2)
┌──────────────────────┐           ┌──────────────────────┐
│ DynamoDB GlobalTable  │──auto────▶│ DynamoDB Replica      │
│ (Source)              │ replicate │ (MultiAccountReplica) │
└──────────────────────┘           └──────────────────────┘
```

### Key Features

- First true multi-account Global Tables pattern in this repo
- Uses `TableV2MultiAccountReplica` (CDK native, no custom resources)
- CDK automatically configures cross-account resource policies
- Two separate stacks deployed to two accounts independently
- `globalTableSettingsReplicationMode: ALL` on both source and replica

### Framework / Language

- AWS CDK (TypeScript)

### Deployment & Testing

- Deployed to two AWS accounts and tested end-to-end
- Write in Account A (us-east-1), read from Account B (us-west-2) within 3 seconds
- Both stacks create/destroy cleanly

### DIFF
--- a/dynamodb-global-tables-multi-account-cdk/.gitignore
+++ b/dynamodb-global-tables-multi-account-cdk/.gitignore
@@ -1,0 +1,6 @@
+*.js
+*.d.ts
+node_modules
+cdk.out
+cdk.context.json
+package-lock.json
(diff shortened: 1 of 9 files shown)


ellisms:
Hi @NithinChandranR-AWS , I reviewed this submission and don't think it warrants a pattern on serverlessland. Since DynamoDB already handles the replication, I don't think there is any added value in a pattern that is just creating the tables.

NithinChandranR-AWS:
Thanks for the review, @ellisms. You're right -- DynamoDB handles replication natively and the pattern doesn't add meaningful value beyond what the service provides out of the box. Closing this one.

PR #3127, opened 2026-06-07
eventbridge-cloudwatch-cdk: Update aws-cdk-lib

kakakakakku:
*Issue #, if available:*
N/A

*Description of changes:*

To keep this pattern maintainable, I updated the aws-cdk-lib.

## Check

`cdk deploy` completed successfully and works good.

```sh
$ aws events put-events --entries file://event.json
{
    "FailedEntryCount": 0,
    "Entries": [
        {
            "EventId": "36769164-46ea-44fb-8980-393787ba17e8"
        }
    ]
}
```

image

Thank you😀

---

### DIFF
--- a/eventbridge-cloudwatch-cdk/README.md
+++ b/eventbridge-cloudwatch-cdk/README.md
@@ -12,7 +12,7 @@ Important: this application uses various AWS services and there are costs associ
 * [AWS CLI](https://docs.aws.amazon.com/cli/latest/userguide/install-cliv2.html) installed and configured
 * [Git Installed](https://git-scm.com/book/en/v2/Getting-Started-Installing-Git)
 * [Node and NPM](https://nodejs.org/en/download/) installed
-* [Python, pip, virtuenv](https://docs.aws.amazon.com/cdk/latest/guide/work-with-cdk-python.html) installed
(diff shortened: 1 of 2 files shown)


parikhudit:
Thanks for bumping the version!


One observation: `aws-cdk-lib==2.258.0` declares `requires_python >=3.10` on PyPI, whereas the old `2.13.0` pin supported much older interpreters. The README's Requirements section only says "Python, pip, virtuenv installed" with no minimum version. A user on `Python 3.9` or earlier could now get a hard pip install failure after this bump. This is a documentation gap surfaced (not caused) by the upgrade. Appreciate if you could cover that.

kakakakakku:
@parikhudit 
Thanks for the comment. I agree with you. Making the prerequisites explicit is helpful for users of this pattern!

Python 3.9 has already reached EOL, so I've updated the README to state `Python 3.10+`, matching the requirement of aws-cdk-lib.
https://devguide.python.org/versions/
https://pypi.org/project/aws-cdk-lib/2.258.0/

For reference, a related pattern **apigw-python-cdk-lambda-snapstart** documents `Python 3.12+` in the same style.
https://github.com/aws-samples/serverless-patterns/tree/main/apigw-python-cdk-lambda-snapstart

Thanks again!

PR #3188, opened 2026-06-24
feat: Add DynamoDB Streams CDC pattern with Lambda durable functions …

awsTest1992:
…(Python)

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/lambda-durable-ddb-streams-python-sam/src/processor.py
+++ b/lambda-durable-ddb-streams-python-sam/src/processor.py
@@ -1,0 +1,172 @@
+"""
+DynamoDB Streams → Lambda Durable Function.
+
+Processes stream records through a multi-step pipeline:
+  1. Validate — check record integrity
+  2. Enrich — add computed fields and metadata
+  3. Notify — write an audit notification record
+
+Each step is checkpointed. If the function is interrupted during processing,
+it resumes from the last checkpoint without re-executing completed steps.
+
+Uses ReportBatchItemFailures for partial batch error handling.
+"""
+
+import os
+import uuid
+import logging
+from datetime import datetime, timezone
+from decimal import Decimal
+
+import boto3
+from aws_durable_execution_sdk_python.context import DurableContext, StepContext, durable_step
+from aws_durable_execution_sdk_python.execution import durable_execution
+
+logger = logging.getLogger()
+logger.setLevel(logging.INFO)
+
+dynamodb = boto3.resource("dynamodb")
+processed_table = dynamodb.Table(os.environ["PROCESSED_TABLE"])
+notifications_table = dynamodb.Table(os.environ["NOTIFICATIONS_TABLE"])
+
+
+@durable_step
+def validate_record(step_context: StepContext, record: dict) -> dict:
+    """Step 1: Validate the stream record has required fields."""
+    step_context.logger.info("Validating record: %s", record.get("pk"))
+
+    event_name = record.get("event_name")
+    new_image = record.get("new_image", {})
+
+    if event_name == "REMOVE":
+        # Deletions are valid but we skip enrichment
+        return {"valid": True, "skip_processing": True, "event_name": event_name}
+
+    # Validate required fields for INSERT/MODIFY
+    required_fields = ["pk", "name"]
+    missing = [f for f in required_fields if f not in new_image]
+
+    if missing:
+        raise UnrecoverableInvocationError(
+            f"Record missing required fields: {missing}"
+        )
+
+    return {
+        "valid": True,
+        "skip_processing": False,
+        "event_name": event_name,
+        "data": new_image,
+    }
+
+
+@durable_step
+def enrich_record(step_context: StepContext, validation: dict) -> dict:
+    """Step 2: Enrich the record with computed fields."""
+    step_context.logger.info("Enriching record: %s", validation["data"].get("pk"))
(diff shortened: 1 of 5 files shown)


lambda-durable-ddb-streams-python-sam/template.yaml
```diff
@@ -0,0 +1,115 @@
...
+            TableName: !Ref NotificationsTable
+        - DynamoDBStreamReadPolicy:
+            TableName: !Ref SourceTable
+            StreamName: !Select [3, !Split ["/", !GetAtt SourceTable.StreamArn]]
+        - Version: '2012-10-17'
+          Statement:
+            - Effect: Allow
+              Action:
+                - lambda:InvokeFunction
+              Resource: !Sub "arn:aws:lambda:${AWS::Region}:${AWS::AccountId}:function:${AWS::StackName}-processor*"
```
ellisms, line 88: Recommend changing resource statement to:
`Resource:
        - !Sub "arn:aws:lambda:${AWS::Region}:${AWS::AccountId}:function:${AWS::StackName}-processor"
        - !Sub "arn:aws:lambda:${AWS::Region}:${AWS::AccountId}:function:${AWS::StackName}-processor:*"
     `

lambda-durable-ddb-streams-python-sam/src/processor.py
```diff
@@ -0,0 +1,172 @@
...
+it resumes from the last checkpoint without re-executing completed steps.
+
+Uses ReportBatchItemFailures for partial batch error handling.
+"""
+
+import os
+import uuid
+import logging
+from datetime import datetime, timezone
+from decimal import Decimal
```
ellisms, line 19: This looks like an unused import. Remove if not needed.

lambda-durable-ddb-streams-python-sam/src/processor.py
```diff
@@ -0,0 +1,172 @@
...
+    if event_name == "REMOVE":
+        # Deletions are valid but we skip enrichment
+        return {"valid": True, "skip_processing": True, "event_name": event_name}
+
+    # Validate required fields for INSERT/MODIFY
+    required_fields = ["pk", "name"]
+    missing = [f for f in required_fields if f not in new_image]
+
+    if missing:
+        raise UnrecoverableInvocationError(
```
ellisms, line 50: I think you need to import this

lambda-durable-ddb-streams-python-sam/template.yaml
```diff
@@ -0,0 +1,115 @@
...
+      BillingMode: PAY_PER_REQUEST
+      AttributeDefinitions:
+        - AttributeName: notification_id
+          AttributeType: S
+      KeySchema:
+        - AttributeName: notification_id
+          KeyType: HASH
+
+  # Durable Function — processes stream records with checkpoints
+  StreamProcessorFunction:
```
ellisms, line 60: Recommend explicitly creating the associated CloudWatch logs group so that there is a retention policy, and the group is removed during stack destroy.

lambda-durable-ddb-streams-python-sam/src/processor.py
```diff
@@ -0,0 +1,172 @@
...
+import uuid
+import logging
+from datetime import datetime, timezone
+from decimal import Decimal
+
+import boto3
+from aws_durable_execution_sdk_python.context import DurableContext, StepContext, durable_step
+from aws_durable_execution_sdk_python.execution import durable_execution
+
+logger = logging.getLogger()
```
ellisms, line 25: Recommend using the `context.logger` from the durable function sdk so that the logger is replay-aware.

ellisms (changes requested):
Thanks for the submission, @awsTest1992 . Requesting a few changes.

awsTest1992:
Thanks for the review @ellisms! I've pushed a commit addressing all the feedback.

**IAM invoke resource (template.yaml)**
Updated to the explicit two-ARN form as recommended:
```yaml
Resource:
  - !Sub "arn:aws:lambda:${AWS::Region}:${AWS::AccountId}:function:${AWS::StackName}-processor"
  - !Sub "arn:aws:lambda:${AWS::Region}:${AWS::AccountId}:function:${AWS::StackName}-processor:*"
```

**Explicit CloudWatch log group (template.yaml)**
Added an `AWS::Logs::LogGroup` (`/aws/lambda/${AWS::StackName}-processor`, `RetentionInDays: 14`) and wired it via `LoggingConfig.LogGroup`, so the group has a retention policy and is removed on stack delete.

**Unused import (processor.py)**
Removed — dropped both `logging` and `Decimal` (neither was used after the logger change below).

**Missing import (processor.py)**
Good catch. The handler raised an exception that wasn't importable as written. I imported `ExecutionError` from the durable SDK and used it for the "missing required fields" case — it returns FAILED without retry, which is the right behavior for a permanent data problem (the record then flows to `batchItemFailures`):
```python
from aws_durable_execution_sdk_python import ExecutionError
```

**Replay-aware logging (processor.py)**
Removed the module-level `logging.getLogger()` and rely on the replay-aware `context.logger` / `step_context.logger` (already used throughout the steps), so log lines aren't duplicated on replay.

**Verification**
`python -m py_compile`, `sam validate --lint`, and `cfn-lint` all pass clean.

Ready for another look — thanks!

PR #3217, opened 2026-07-07
Strands agent with AgentCore Gateway with MCP target on Lambda (SAM)

humem001:
*Issue #, if available:*

*Description of changes:*


strands-agentcore-mcp/infrastructure/.gitkeep
bfreiberg, line 1: Does need to be kept?

strands-agentcore-mcp/example-pattern.json
```diff
@@ -0,0 +1,86 @@
+{
+  "title": "Serverless AI Agent with AgentCore MCP Gateway on Lambda",
```
bfreiberg:
```suggestion
  "title": "Agent using MCP server on AWS Lambda through Amazon Bedrock AgentCore Gateway",
```

strands-agentcore-mcp/example-pattern.json
```diff
@@ -0,0 +1,86 @@
...
+    "text": [
+      "./scripts/test.sh",
+      "./scripts/test.sh 'List all products in Electronics'",
+      "./scripts/test.sh 'Get product ELEC-001 details'",
+      "./scripts/test.sh 'Add a new product called Widget Pro in Electronics for $49.99'"
+    ]
+  },
+  "cleanup": {
+    "text": [
+      "./scripts/terminate.sh"
```
bfreiberg:
```suggestion
      "<code>./scripts/terminate.sh</code>"
```

strands-agentcore-mcp/infrastructure/template.yaml
```diff
@@ -0,0 +1,450 @@
...
+
+# ---------------------------------------------------------------------------
+# Resources
+# ---------------------------------------------------------------------------
+# NOTE: This section is authored incrementally across subtasks 1.1–1.4.
+#   1.1 (this subtask): Cognito User Pool + Client, DynamoDB Product Table
+#   1.2: IAM roles (AgentLambdaRole, McpServerRole, GatewayExecutionRole)
+#   1.3: Serverless::Function definitions (AgentLambdaFunction, McpServerLambda)
+#   1.4: API Gateway resources + AgentCoreGateway
+# ---------------------------------------------------------------------------
```
bfreiberg: Remove

strands-agentcore-mcp/infrastructure/template.yaml
```diff
@@ -0,0 +1,450 @@
...
+          PolicyDocument:
+            Version: '2012-10-17'
+            Statement:
+              - Effect: Allow
+                Action:
+                  - dynamodb:GetItem
+                  - dynamodb:PutItem
+                  - dynamodb:Query
+                  - dynamodb:Scan
+                # Scoped to ProductTable ARN ONLY — no wildcard (Requirement 9.2, 9.4)
```
bfreiberg:
```suggestion
                # Scoped to ProductTable ARN ONLY 
```

bfreiberg: Remove these self-justifying comments please

strands-agentcore-mcp/infrastructure/template.yaml
```diff
@@ -0,0 +1,450 @@
...
+  # =========================================================================
+  #
+  # CASING TRAPS (from project-conventions.md):
+  #   - AgentCoreGateway uses RoleArn, NOT ExecutionRoleArn
+  #   - AuthorizerConfiguration uses CustomJWTAuthorizer (JWT all-caps)
+  #   - AllowedAudience, NOT Audience
+  #   - CredentialProviderConfigurations is an ARRAY
+  #   - DiscoveryUrl must end with /.well-known/openid-configuration
+  # =========================================================================
+
```
bfreiberg: remove

bfreiberg: Please apply this throughout the file

strands-agentcore-mcp/infrastructure/template.yaml
```diff
@@ -0,0 +1,450 @@
...
+                  - !Sub 'arn:aws:bedrock-agentcore:${AWS::Region}:${AWS::AccountId}:workload-identity-directory/default'
+                  - !Sub 'arn:aws:bedrock-agentcore:${AWS::Region}:${AWS::AccountId}:workload-identity-directory/default/workload-identity/${GatewayName}-*'
+        - PolicyName: InvokeMcpApi
+          PolicyDocument:
+            Version: '2012-10-17'
+            Statement:
+              - Effect: Allow
+                Action: execute-api:Invoke
+                # Scoped to all APIs in this account/region — the API Gateway resource
+                # policy on McpApi further restricts this to the specific API.
```
bfreiberg: I do not see a resource policy on the McpApi. Is it missing?

strands-agentcore-mcp/infrastructure/template.yaml
```diff
@@ -0,0 +1,450 @@
...
+
+  # =========================================================================
+  # Agent Lambda — function (AWS::Serverless::Function, Docker-free makefile build)
+  # =========================================================================
+
+  AgentLambdaFunction:
+    Type: AWS::Serverless::Function
+    Properties:
+      FunctionName: !Sub '${AWS::StackName}-agent'
+      Runtime: python3.12
```
bfreiberg: use latest runtime version please

strands-agentcore-mcp/infrastructure/template.yaml
```diff
@@ -0,0 +1,450 @@
...
+
+  # =========================================================================
+  # MCP Server Lambda — function (AWS::Serverless::Function, Docker-free makefile build)
+  # =========================================================================
+
+  McpServerLambda:
+    Type: AWS::Serverless::Function
+    Properties:
+      FunctionName: !Sub '${AWS::StackName}-mcp-server'
+      Runtime: python3.12
```
bfreiberg: see above

strands-agentcore-mcp/infrastructure/template.yaml
```diff
@@ -0,0 +1,450 @@
...
+      RestApiId: !Ref McpApi
+      ResourceId: !Ref McpResource
+      HttpMethod: POST
+      # NONE auth — AgentCore Gateway signs requests with SigV4 (execute-api service)
+      # using the GatewayExecutionRole. The gateway's IAM policy + the role's
+      # execute-api:Invoke permission provide the access control layer.
+      # AWS_IAM auth on the method causes 403 because API GW rejects cross-service
+      # assumed-role SigV4 calls without an explicit resource policy allowing the
+      # assumed-role session ARN, which cannot be expressed cleanly in CloudFormation.
+      AuthorizationType: NONE
```
bfreiberg: I would except a resource policy here (or AuthorizationType: IAM)

strands-agentcore-mcp/scripts/deploy.sh
```diff
@@ -0,0 +1,501 @@
...
+# Usage:
+#   ./scripts/deploy.sh
+#
+# Prerequisites: AWS SAM CLI + pip3 (Python 3.12). No Docker required —
+# the build is Docker-free via the root Makefile's BuildMethod: makefile
+# targets, which run the two-step pip3 manylinux install during `sam build`.
+#
+# What it does (in order):
+#   1. Build the SAM application (sam build — Docker-free, Makefile-driven)
+#   2. Deploy the SAM application (sam deploy — reads samconfig.toml)
```
bfreiberg: there is no samconfig.toml in the repo?

strands-agentcore-mcp/src/shared/jwt_utils.py
```diff
@@ -0,0 +1,151 @@
...
+    Args:
+        claims: Decoded JWT claims dictionary.
+
+    Returns:
+        The username string from the cognito:username claim.
+
+    Raises:
+        JWTValidationError: If the cognito:username claim is missing.
+    """
+    username = claims.get("cognito:username")
```
bfreiberg: Cognito access tokens carry username, not cognito:username, so any access token that passes validation then fails at username extraction. Either restrict to ID tokens or fall back to username for access tokens.

strands-agentcore-mcp/src/shared/logging_utils.py
```diff
@@ -0,0 +1,74 @@
...
+        The configured root logger.
+    """
+    log_format = "[%(levelname)s]"
+    if correlation_id:
+        log_format += f" [{correlation_id}]"
+    log_format += " %(name)s - %(message)s"
+
+    logging.basicConfig(level=level, format=log_format, force=True)
+    # Enable debug logging for strands to see tool call details
+    logging.getLogger("strands").setLevel(logging.DEBUG)
```
bfreiberg: Strands debug logs tool-call payloads and HTTP request details, which can include the Authorization: Bearer <JWT> header and full prompts/results. Bearer tokens and user content in CloudWatch is a data-exposure risk

bfreiberg (changes requested):

strands-agentcore-mcp/README.md
```diff
@@ -0,0 +1,161 @@
+# strands-agentcore-mcp
```
bfreiberg:
```suggestion
# Agent using MCP server on AWS Lambda through Amazon Bedrock AgentCore Gateway
```

strands-agentcore-mcp/README.md
```diff
@@ -0,0 +1,161 @@
...
+
+A serverless AI agent that answers natural-language prompts about products by invoking tools through AWS Bedrock AgentCore Gateway using the Model Context Protocol (MCP).
```
bfreiberg:
```suggestion
A serverless AI agent that answers natural-language prompts about products by invoking tools through Amazon Bedrock AgentCore Gateway using the Model Context Protocol (MCP).
```

strands-agentcore-mcp/README.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Architecture
+
+![Architecture Diagram](architecture/mcp-target.png)
+
+```
+User → Agent Lambda → AgentCore Gateway (MCP) → API Gateway → MCP Server Lambda → DynamoDB
```
bfreiberg:
```suggestion
User → Agent Lambda function → AgentCore Gateway (MCP) → API Gateway → MCP Server Lambda function → DynamoDB
```

strands-agentcore-mcp/README.md
```diff
@@ -0,0 +1,161 @@
...
+
+### Step 3: Deploy
+
+```bash
+./scripts/deploy.sh
+```
+
+To use a different model, edit the `BedrockModelId` value in the `parameter_overrides` line of `samconfig.toml` before running (or pass `sam deploy --parameter-overrides 'BedrockModelId="<model-id>"'`).
+
+The deploy script is a thin wrapper around AWS SAM. It runs `sam build` (Docker-free — no `--use-container`) followed by `sam deploy` (which reads `samconfig.toml`), then performs the post-deploy steps. In order it will:
```
bfreiberg:
```suggestion
The deploy script is a thin wrapper around AWS SAM. It runs `sam build` followed by `sam deploy`, then performs the post-deploy steps. In order it will:
```

strands-agentcore-mcp/README.md
```diff
@@ -0,0 +1,161 @@
...
+
+1. Build the SAM application with `sam build` — the native `BuildMethod: makefile` build (driven by the root `Makefile`) installs Lambda dependencies via the two-step `pip3` manylinux install and packages the real `src/` tree, no Docker needed
+2. Deploy the stack with `sam deploy` — Cognito, DynamoDB, API Gateway, Lambda x2, AgentCore Gateway, and IAM roles are created (or updated) in one shot with the real Lambda code, no inline placeholders
```
bfreiberg:
```suggestion
2. Deploy the stack with `sam deploy` — Cognito, DynamoDB, API Gateway, Lambda x2, AgentCore Gateway, and IAM roles are created (or updated) 
```

strands-agentcore-mcp/README.md
```diff
@@ -0,0 +1,161 @@
...
+3. Read the stack outputs
+4. Create and synchronize the MCP Gateway Target
+5. Seed DynamoDB with 3 sample products
```
bfreiberg:
```suggestion
5. Seed the DynamoDB table with 3 sample products
```

strands-agentcore-mcp/README.md
```diff
@@ -0,0 +1,161 @@
...
+6. Create a Cognito test user (`testuser` / `TestPass123!`)
+7. Generate `scripts/test.sh` with deployment values baked in
+
+Deploy takes approximately 5–10 minutes on first run.
+
+> **Packaging convention.** The native SAM `BuildMethod: makefile` build replaces the old in-script two-step `pip3` packaging and the separate `aws lambda update-function-code` flow. The proven two-step `pip3 --platform manylinux2014_x86_64 --only-binary=:all:` install (binary packages, then pure-Python packages `--no-deps`) now lives in the root `Makefile`'s `build-AgentLambdaFunction` / `build-McpServerLambda` targets and runs during `sam build`. `sam deploy` ships the real code directly, so there is no manual zip-and-update step.
```
bfreiberg: remove

strands-agentcore-mcp/README.md
```diff
@@ -0,0 +1,161 @@
...
+```
+
+## Running Tests
+
+```bash
+pip3 install -r requirements-dev.txt
+python3 -m pytest tests/unit/ tests/property/ -v
+```
+
+## Notes
```
bfreiberg: Remove if already stated in the doc

bfreiberg (changes requested):

strands-agentcore-mcp/infrastructure/.gitkeep
humem001: .gitkeep files removed

strands-agentcore-mcp/example-pattern.json
humem001: Title updated

strands-agentcore-mcp/example-pattern.json
humem001: Applied <code> tags consistentl

humem001: y

strands-agentcore-mcp/infrastructure/template.yaml
humem001: removed

strands-agentcore-mcp/infrastructure/template.yaml
humem001: references removed from the template

strands-agentcore-mcp/infrastructure/template.yaml
humem001: The CASING TRAPS block is gone.

strands-agentcore-mcp/infrastructure/template.yaml
humem001: execute-api:Invoke is now scoped to ${McpApi}/*/POST/mcp — the specific API, method, and path — instead of the broad */*/*/*. The misleading comment about a non-existent resource policy is gone. There was no circular dependency preventing this; the old comment was incorrect.

McpMethod comment updated to accurately describe what actually enforces access control — the scoped execute-api:Invoke grant in GatewayExecutionRole.

strands-agentcore-mcp/infrastructure/template.yaml
humem001: done

strands-agentcore-mcp/infrastructure/template.yaml
humem001: done

strands-agentcore-mcp/infrastructure/template.yaml
humem001: McpMethod.AuthorizationType changed from NONE → AWS_IAM. API Gateway now validates the SigV4 signature and enforces that the caller has execute-api:Invoke on this endpoint. The GatewayExecutionRole already has exactly that permission scoped to POST /mcp — so AgentCore Gateway can invoke it and nothing else can.
Unit test renamed test_mcp_method_auth_type_none → test_mcp_method_auth_type and updated to assert AWS_IAM.

strands-agentcore-mcp/scripts/deploy.sh
humem001: looks like I had not committed it.. should be there now

strands-agentcore-mcp/src/shared/jwt_utils.py
humem001: Here's what changed:

extract_username now tries cognito:username first (ID token), then falls back to username (access token). Both validate_token accepts — now both will also succeed at username extraction.

strands-agentcore-mcp/src/shared/logging_utils.py
humem001: The strands DEBUG override is removed.

strands-agentcore-mcp/README.md
humem001: done

strands-agentcore-mcp/README.md
humem001: done

strands-agentcore-mcp/README.md
humem001: done

strands-agentcore-mcp/README.md
humem001: done

strands-agentcore-mcp/README.md
humem001: done

strands-agentcore-mcp/README.md
humem001: done

strands-agentcore-mcp/README.md
humem001: done

strands-agentcore-mcp/README.md
humem001: done

strands-agentcore-mcp/infrastructure/template.yaml
humem001: We investigated this during a live deploy test. We switched AuthorizationType to AWS_IAM and the MCP target synchronization immediately failed with 502 errors — AgentCore's backend probes tools/list during target creation/sync using a service-level assumed role that differs from GatewayExecutionRole, so API Gateway's IAM authorizer rejected the probe.

AuthorizationType: NONE is required for AgentCore MCP targets to function. The security model is layered differently from a standard API: the AgentCore Gateway's CUSTOM_JWT authorizer validates the Cognito JWT on every user request before it reaches the MCP endpoint, so unauthenticated direct access to /prod/mcp would receive a JSON-RPC handler response but would never result in an authenticated agent session. A resource policy restricting the caller to GatewayExecutionRole would also block the AgentCore probe for the same reason.

We've updated the comment in the template to clearly explain this constraint so it's not mistaken for a missing security control.

The template comment currently reads:


# AuthorizationType NONE — AgentCore Gateway signs requests with SigV4
# using GatewayExecutionRole, but API Gateway's AWS_IAM authorizer rejects
# cross-service assumed-role SigV4 calls from AgentCore. Access is controlled
# by the GatewayExecutionRole's execute-api:Invoke grant scoped to this API.
AuthorizationType: NONE
That's accurate and explains the tradeoff clearly for anyone who reads the template in the future

bfreiberg: I deployed this pattern to a test account and validated the behavior end-to-end, because I wanted to be sure before we settle the thread. Sharing the results since they change the conclusion.

**Baseline (as merged — `AuthorizationType: NONE`, no resource policy) is exploitable by an anonymous caller.** A plain `curl` to `/prod/mcp` with **no `Authorization` header**:

```
# tools/list             -> HTTP 200, full catalog
# tools/call put_product -> HTTP 200 {"written": true}
```

The written item then showed up in DynamoDB. So "unauthenticated direct access to `/prod/mcp` would receive a JSON-RPC handler response but would never result in an authenticated agent session" is the vulnerability, not a mitigation — the attacker doesn't need an agent session. The MCP Lambda parses the request, dispatches the tool, and **writes to the table** for anyone on the internet. The `CUSTOM_JWT` authorizer only sits in front of the *gateway* path; a direct call to the API Gateway URL never traverses the gateway, so the JWT is never checked.

**On "AWS_IAM breaks AgentCore sync":** I couldn't reproduce that. What I *did* find is that the fix has two parts, and method auth alone or a resource policy alone is not enough:

- Resource policy **with the method still `NONE`** → anonymous requests carry no `aws:PrincipalArn`, so the principal condition never engages. Anonymous `put_product` still returned `HTTP 200` and wrote the item. (So a resource policy by itself does not close this — worth flagging.)
- `AuthorizationType: AWS_IAM` on the method **+** a resource policy allowing only `GatewayExecutionRole`:
  - Anonymous `curl` → `HTTP 403` *"User: anonymous is not authorized … explicit deny in a resource-based policy"*, and nothing written to DynamoDB.
  - **`synchronize_gateway_targets` → `READY`.** The AgentCore probe is signed by `GatewayExecutionRole` (the `GATEWAY_IAM_ROLE` / `execute-api` credential provider configured in `deploy.sh`), so it passes both the IAM authorizer and the resource policy.

So the endpoint can be locked to the gateway role while AgentCore continues to work. The earlier 502 was likely method `AWS_IAM` without the accompanying resource-policy `Allow` for the role (or a missing stage redeploy), rather than a fundamental incompatibility.

Concretely:

```yaml
McpMethod:
  Properties:
    AuthorizationType: AWS_IAM        # was NONE

McpApi:
  Properties:
    Name: mcp-api
    Policy:
      Version: '2012-10-17'
      Statement:
        - Effect: Allow
          Principal: { AWS: !GetAtt GatewayExecutionRole.Arn }
          Action: execute-api:Invoke
          Resource: !Sub 'arn:aws:execute-api:${AWS::Region}:${AWS::AccountId}:*/*/POST/mcp'
```

The circular-reference concern is avoidable: keep the role's `InvokeMcpApi` grant scoped with `!Sub` (not `!GetAtt McpApi`) and reference the role from the API policy one-directionally.

Given this is a sample others will copy, and the default currently exposes a write-capable (`put_product`) endpoint to the internet, I think we should land the `AWS_IAM` + resource-policy change rather than resolve as working-as-intended. Happy to push the commit if that's easier.

strands-agentcore-mcp/scripts/deploy.sh
bfreiberg: I do not see it so far

strands-agentcore-mcp/scripts/deploy.sh
```diff
@@ -0,0 +1,501 @@
...
+# ---------------------------------------------------------------------------
+# STEP 2: Deploy the SAM application (Requirement 10.1, 10.2)
+#
+# All deploy parameters (stack name, region, capabilities, resolve_s3,
+# parameter_overrides, fail_on_empty_changeset=false) live in samconfig.toml,
+# so no flags are passed here. SAM handles create-vs-update automatically and
+# tolerates no-op changesets.
+# ---------------------------------------------------------------------------
+log "Step 2: Deploying SAM application (stack '$STACK_NAME')..."
+sam deploy
```
bfreiberg: runs bare sam deploy ("reads samconfig.toml"), samconfig.toml is still git-ignored (.gitignore:43) and not in the repo, and README (lines 53, 115) still references it as present. A fresh clone still can't deploy as documented. Commit a sample samconfig.toml or pass the flags explicitly.

bfreiberg (changes requested):

strands-agentcore-mcp/scripts/deploy.sh
humem001: samconfig.toml has been removed from .gitignore and is now committed to the repo. It contains only non-sensitive values (stack name agentcore-mcp, region us-east-1, CAPABILITY_NAMED_IAM, resolve_s3 = true, and the default BedrockModelId — no account IDs or credentials). A fresh clone can now run 
deploy.sh
 without any manual setup. The README has also been updated to note that region and model ID can be changed via samconfig.toml before deploying.

The deploy.sh header comment had a stale Python 3.14 reference — fixed to 3.13, consistent with the Makefile, template, and README. Also caught and fixed the same stale reference in the project conventions doc. All version references now consistently say python3.13.

strands-agentcore-mcp/infrastructure/template.yaml
humem001: Fixed. The vulnerability was real — anonymous callers could POST directly to /prod/mcp and write to DynamoDB without going through the AgentCore Gateway's JWT auth. Now fixed with AuthorizationType: AWS_IAM on the method plus an API Gateway resource policy.

The resource policy uses Principal: '*' with Condition: ArnLike: aws:PrincipalArn: <role-arn> rather than a named Principal, which avoids the CloudFormation circular dependency (API Gateway validates named Principals exist at creation time, but the role and API are created in the same changeset). The effect is identical: only signed requests from GatewayExecutionRole are allowed; anonymous curl now returns HTTP 403. MCP target synchronization still reaches READY. Verified end-to-end.

bfreiberg:
Looks good, thank you for your contribution. Your pattern will be merged soon.

bfreiberg (approved):

PR #3224, opened 2026-07-10
Checked in new pattern rabbitmq-private-lambda-python-sam

gmedard:
*Issue #, if available:*

*Description of changes:* - We created a new pattern rabbitmq-private-lambda-java-sam a few months ago. Now checking in the Python equivalent of the pattern rabbitmq-private-lambda-python-sam

When running the CloudFormation template, please remember to change the input parameter ServerlessLandGithubLocation from https://github.com/aws-samples/serverless-patterns.git to https://github.com/gmedard/serverless-patterns.git, else the sample will not work. This is only for testing till the PR is merged. Customers won't have to undertake this step.


rabbitmq-private-lambda-python-sam/README.md
```diff
@@ -0,0 +1,180 @@
...
+
+## Run the AWS CloudFormation template to create the Amazon MQ (RabbitMQ) Cluster and Client EC2 instance
+
+* [Run the AWS CloudFormation template using the file RabbitMQAndClientEC2.yaml] - You can go to the AWS CloudFormation console, create a new stack by specifying the template file. You can keep the defaults for input parameters or modify them as necessary. Wait for the AWS CloudFormation stack to be created. This AWS CloudFormation template will create an Amazon MQ (RabbitMQ) cluster. It will also create an EC2 instance that you can use as a client.
+
+* [Connect to the EC2 instance] - Once the AWS CloudFormation stack is created, you can go to the EC2 console and log into the instance using either "Connect using EC2 Instance Connect" or "Connect using EC2 Instance Connect Endpoint" option under the "EC2 Instance Connect" tab. In case you are using SSM Instance connect, you are not initially placed in the home directory. If you connect as ssm-user, you need to sudo su to ec2-user for this to work.
```
bfreiberg:
```suggestion
* [Connect to the EC2 instance] - Once the AWS CloudFormation stack is created, you can go to the EC2 console and log into the instance using either "Connect using EC2 Instance Connect" or "Connect using EC2 Instance Connect Endpoint" option under the "EC2 Instance Connect" tab. In case you are using SSM Instance connect, you are not initially placed in the home directory. If you connect as ssm-user, you need to `sudo su` to `ec2-user` for this to work.
```

rabbitmq-private-lambda-python-sam/README.md
```diff
@@ -0,0 +1,180 @@
...
+Run functions locally and invoke them with the `sam local invoke` command.
+
+```bash
+cd /home/ec2-user/serverless-patterns/rabbitmq-private-lambda-python-sam/rabbitmq_consumer_dynamo_sam
+sam local invoke --event events/event.json
+```
+
+## Check RabbitMQ Queue Creation
+
+We have included a shell script file called create_rabbit_queue.sh. This shell script creates a VirtualHost, an Exchange and a Queue in the RabbitMQ cluster. It also binds the queue to the exchange. This step is necessary before the Lambda function can be deployed.
```
bfreiberg:
```suggestion
We have included a shell script file called `create_rabbit_queue.sh`. This shell script creates a VirtualHost, an Exchange and a Queue in the RabbitMQ cluster. It also binds the queue to the exchange. This step is necessary before the Lambda function can be deployed.
```

rabbitmq-private-lambda-python-sam/README.md
```diff
@@ -0,0 +1,180 @@
...
+
+The create_rabbit_queue.sh script is run automatically as part of the AWS CloudFormation script.
```
bfreiberg:
```suggestion
The `create_rabbit_queue.sh` script is run automatically as part of the AWS CloudFormation script.
```

rabbitmq-private-lambda-python-sam/README.md
```diff
@@ -0,0 +1,180 @@
...
+
+There is another shell script file called query_rabbit_queue.sh that has been included. You can run this script to ensure the virtualhost, exchange and queue have been created in the RabbitMQ cluster:
```
bfreiberg:
```suggestion
There is another shell script file called `query_rabbit_queue.sh` that has been included. You can run this script to ensure the virtualhost, exchange and queue have been created in the RabbitMQ cluster:
```

rabbitmq-private-lambda-python-sam/README.md
```diff
@@ -0,0 +1,180 @@
...
+Sent out one message - Number 6 at time = 1760937987914
+Sent out one message - Number 7 at time = 1760937987915
+Sent out one message - Number 8 at time = 1760937987916
+Sent out one message - Number 9 at time = 1760937987918
+Sent out one message - Number 10 at time = 1760937987919
+
+
+Once the messages have been sent, check Amazon CloudWatch logs and you should see messages for the Amazon CloudWatch Log Group with the name of the deployed Lambda function.
+
+When you run the above script, it sends messages with JSON records to the Amazon MQ (RabbitMQ) cluster on the queue on which the lambda function is listening on. The Lambda function listens on the published Amazon MQ (RabbitMQ) messages on the queue.
```
bfreiberg:
```suggestion
When you run the above script, it sends messages with JSON records to the Amazon MQ (RabbitMQ) cluster on the queue on which the Lambda function is listening on.
```

rabbitmq-private-lambda-python-sam/README.md
bfreiberg, line 1: add copyright footer to the end of the file

bfreiberg (changes requested):
## Code review — `rabbitmq-private-lambda-python-sam`

Thanks for the contribution! I reviewed the pattern and validated the findings against a live end-to-end deployment. Findings below are grouped by severity, and I've marked each with how it was verified.

### 🔴 High — data loss

**1. Consumer silently drops the entire batch on any error** — `rabbitmq_consumer_dynamo_sam/rabbitmq_event_consumer_function/lambda_function.py:19-90`

`lambda_handler` wraps the whole message loop in `try/except` and, on any exception, logs it and `return "500"`. Amazon MQ event source mappings ignore the handler's return value and only retry/redrive when the invocation **raises**. Returning normally is treated as success, so the batch is acked and removed from the queue.

Verified live: I pointed the function at a table it couldn't write, published 20 messages, and observed:
- ESM delivered 2 batches; each logged `An exception occurred - ... PutItem ... AccessDeniedException`
- Lambda `Errors` metric = **0**, `Invocations` = 2 (ESM saw success)
- DynamoDB table count = **0** (nothing persisted)
- Queue depth = **0**, no redelivery — the messages were gone

Because the `try` wraps the whole loop, the logs also showed only the *first* message of each batch reached `put_item` before the exception aborted the remaining 9 — so one bad/throttled message discards everything after it in the batch too.

Suggested fix: let exceptions propagate (remove the catch-all), or adopt partial-batch responses (`ReportBatchItemFailures` / `bisectBatchOnFunctionError` + a DLQ) so failed messages are retried instead of silently dropped.

**2. Producer off-by-one skips the first record and crashes at the max count** — `rabbitmq_message_sender_json/rabbitmq_producer.py:76-77`

`for i in range(1, num_to_send + 1): ... people[i]` skips `people[0]` and reads out of bounds when `num_to_send == len(people)`. `us-500.csv` has 500 rows (0–499, no header).

Verified live: running the real producer with `500` raised `IndexError: list index out of range` at `people[500]` after publishing only 499 messages (the first CSV record was never sent). The README explicitly tells users to pass "a number between 1 and 500", so `500` is a documented input.

Suggested fix: `for i in range(num_to_send): person = get_person_from_row(people[i])` (and update the header/number fields accordingly).

### 🟠 Medium — security posture

**3. Broker security group opens all ports to the whole VPC** — `RabbitMQAndClientEC2.yaml:337-340`

Alongside the scoped 5671/443 rules, `RabbitMQSecurityGroup` allows `tcp 0-65535` from the VPC CIDR `10.2.0.0/16`. Verified on the deployed SG. Recommend dropping this rule and keeping only 5671 (and 443 if the console/management API is needed) from the client SG.

**4. Client EC2 instance role is effectively account administrator** — `RabbitMQAndClientEC2.yaml:681-732`

The role attaches `IAMFullAccess`, `AWSLambda_FullAccess`, `AmazonS3FullAccess`, `AmazonDynamoDBFullAccess_v2`, `AWSCloudFormationFullAccess`, `CloudWatchFullAccess`, plus inline `iam:*` and `mq:*` on `*`. Verified on the deployed role. `IAMFullAccess` + `iam:*` also allows privilege escalation. Recommend scoping to the specific actions the setup scripts and `sam deploy` actually need.

**5. Hardcoded default broker password** — `RabbitMQAndClientEC2.yaml:39-43`

`RabbitMQBrokerPassword` defaults to `rabbitmqPassword123`, committed to a public repo and also written into Secrets Manager. Recommend removing the default and requiring the value at deploy time (or generating one via Secrets Manager).

### 🟡 Low — correctness / docs

**6. Empty `messageId` breaks the DynamoDB write** — `lambda_function.py:95`

`MessageID` (the table's partition key) defaults to `''` when `basicProperties.messageId` is absent. Verified live: `PutItem` with an empty-string key returns `ValidationException: The AttributeValue for a key attribute cannot contain an empty string value` (and that error is then swallowed by finding #1). Recommend skipping/dead-lettering messages without a usable key.

**7. `MessageType` stores the message ID, not the type** — `lambda_function.py:116`

`'MessageType': message_id` — should likely be `basic_properties.get('type', '')` (the type is otherwise only stored under `Type`).

**8. `templateFile` points at a non-existent directory** — `example-pattern.json:20`

`"activemq_consumer_dynamo_sam/template_original.yaml"` — this pattern's directory is `rabbitmq_consumer_dynamo_sam` (looks like a leftover from an ActiveMQ pattern). Breaks the generated template link.

**9. `Python3Version` parameter description contradicts its allowed values** — `RabbitMQAndClientEC2.yaml:10-18`

The description says "between 3.9 and 3.12 … maximum allowed version is 3.12", but `AllowedValues` goes up to `python3.14` and the default is `python3.14`.

rabbitmq-private-lambda-python-sam/README.md
gmedard: Applied: `sudo su` to `ec2-user`

rabbitmq-private-lambda-python-sam/README.md
gmedard: Applied: `create_rabbit_queue.sh`

rabbitmq-private-lambda-python-sam/README.md
gmedard: Applied: `create_rabbit_queue.sh`

rabbitmq-private-lambda-python-sam/README.md
gmedard: Applied: `query_rabbit_queue.sh`

rabbitmq-private-lambda-python-sam/README.md
gmedard: Removed "The Lambda function listens on the published Amazon MQ (RabbitMQ) messages on the queue." (duplicated the preceding sentence)

rabbitmq-private-lambda-python-sam/README.md
gmedard: Added standard Copyright 2025 Amazon.com, Inc. ... SPDX-License-Identifier: MIT-0 footer

gmedard:
🔴 High — Data Loss
### 1: Consumer silently drops the entire batch on any error
What we changed: Removed the catch-all try/except block entirely from lambda_handler. Exceptions now propagate naturally, causing the Lambda invocation to fail. The ESM sees the failure and retries the batch (or sends to DLQ if configured).

Why: The ESM contract for Amazon MQ is that a successful invocation (no exception raised) means the batch was processed. The only way to signal failure is to let the exception propagate. This is different from SQS-based event sources where you can use batchItemFailures in the response.

### 2: Producer off-by-one skips the first record and crashes at the max count
What we changed: Changed to for i in range(num_to_send): person = get_person_from_row(people[i]). Updated MessageNumberInBatch, correlation_id, message_id, and the print statement to use i + 1 for 1-based display numbers.

🟠 Medium — Security Posture
### 3: Broker security group opens all ports to the whole VPC
What we changed: Removed the tcp 0-65535 rule entirely. The broker SG now only allows:

Port 5671 from the client SG (AMQPS connections)
Port 5671 from itself (self-referencing — needed for Lambda ESM ENIs)
Port 443 from the client SG (management API/console)

### 4: Client EC2 instance role is effectively account administrator
**What we changed:** Removed all overly broad managed policies and the `iam:*` inline policy. Replaced with a single `SAMDeployPolicy` scoped to specific actions needed for `sam deploy`:
- `cloudformation:*` (stack CRUD, changesets, template operations including `GetTemplateSummary`)
- `lambda:*` (function and event source mapping CRUD)
- `iam:CreateRole/DeleteRole/GetRole/PassRole/AttachRolePolicy/DetachRolePolicy/PutRolePolicy/DeleteRolePolicy/GetRolePolicy/TagRole/UntagRole` scoped to account
- `s3:*` (artifact bucket operations)
- `dynamodb:CreateTable/DeleteTable/DescribeTable/UpdateTable/TagResource/UntagResource`
- `logs:CreateLogGroup/CreateLogStream/PutLogEvents/DescribeLogGroups/DescribeLogStreams`
- `ec2:Describe*` (networking for Lambda VPC config)
- `secretsmanager:GetSecretValue/PutSecretValue/DescribeSecret`

Kept only `AmazonMQFullAccess` (legitimate — scripts call `aws mq describe-broker`) and `AmazonSSMManagedInstanceCore` (for Session Manager access).

### 5: Hardcoded default broker password
**What we changed:** Removed the `RabbitMQBrokerPassword` parameter entirely. The `RabbitMQSecret` resource now uses `GenerateSecretString` to auto-generate a 24-character alphanumeric password at stack creation time:

```yaml
GenerateSecretString:
  SecretStringTemplate: !Sub '{"username": "${RabbitMQBrokerAdminUser}"}'
  GenerateStringKey: 'password'
  PasswordLength: 24
  ExcludePunctuation: true
```

The broker retrieves the password via a CloudFormation dynamic reference:
```yaml
Password: !Sub '{{resolve:secretsmanager:${RabbitMQSecret}:SecretString:password}}'
```

Shell scripts (`create_rabbit_queue.sh`, `query_rabbit_queue.sh`) fetch credentials from Secrets Manager at runtime using `aws secretsmanager get-secret-value`.

## 🟡 Low — Correctness / Docs

### 6: Empty `messageId` breaks the DynamoDB write
**What we changed:** Added a guard at the top of `insert_into_dynamodb`:
```python
if not message_id:
    print("WARNING: Skipping message with empty messageId - cannot use as DynamoDB partition key")
    return
```

### 7: `MessageType` stores the message ID, not the type
**What we changed:** Changed to `'MessageType': basic_properties.get('type', '')`.

### 8: `templateFile` points at a non-existent directory
**What we changed:** Fixed to `"rabbitmq_consumer_dynamo_sam/template_original.yaml"`.

### 9: `Python3Version` parameter description contradicts its allowed values
**What we changed:** Updated description to: "Choose the version of Python 3. Available options are 3.11, 3.12, 3.13, and 3.14. Note that Python 3.10 is not available on Amazon Linux 2023 so it is not offered as an option."

rabbitmq-private-lambda-python-sam/RabbitMQAndClientEC2.yaml
```diff
@@ -0,0 +1,899 @@
...
+    Type: AWS::EC2::SecurityGroup
+    Properties:
+      GroupDescription: Enable SSH access via port 22 from BastionHostSecurityGroup
+      GroupName: !Sub "${AWS::StackName} Security group attached to the RabbitMQ client"
+      VpcId: !Ref VPC
+      SecurityGroupIngress:
+      - IpProtocol: tcp
+        FromPort: 22
+        ToPort: 22
+        CidrIp: 10.0.0.0/16
```
bfreiberg: the VPC CIDR is 10.2.0.0/16 so this range contains no hosts; consider removing it.

rabbitmq-private-lambda-python-sam/RabbitMQAndClientEC2.yaml
```diff
@@ -0,0 +1,899 @@
...
+            sed -i "s|RABBITMQ_BROKER_ENDPOINT|${rabbitmq_broker_endpoint}|g" query_rabbit_queue.sh
+            sed -i "s/RABBITMQ_QUEUE_NAME/$RABBITMQ_QUEUE_NAME/g" query_rabbit_queue.sh
+            sed -i "s/RABBITMQ_VIRTUAL_HOST/$RABBITMQ_VIRTUAL_HOST/g" query_rabbit_queue.sh
+            sed -i "s/RABBITMQ_EXCHANGE/$RABBITMQ_EXCHANGE/g" query_rabbit_queue.sh
+
+            # Get IP CIDR range for EC2 Instance Connect
+            cd /home/ec2-user
+            mkdir -p ip_prefix
+            cd ip_prefix
+            git clone https://github.com/joetek/aws-ip-ranges-json.git
```
bfreiberg: Why use a third-party if the source if AWS owned: https://ip-ranges.amazonaws.com/ip-ranges.json

rabbitmq-private-lambda-python-sam/README.md
```diff
@@ -0,0 +1,185 @@
...
+sam delete
+
+```
+Confirm the delete by selecting Y at both prompts.
+AWS SAM confirms the stack deletion with the "Deleted successfully" message in the terminal.
+
+Next you need to delete the Cloudformation template that created the Amazon MQ (RabbitMQ) cluster and the EC2 instance. Open the Amazon CloudFormation console, select the stack, then choose Delete. The delete will take some time.
+
+----
+Copyright 2025 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
bfreiberg:
```suggestion
Copyright 2026 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```

rabbitmq-private-lambda-python-sam/example-pattern.json
```diff
@@ -0,0 +1,80 @@
...
+      "name": "Angelo Spagnolo",
+      "bio": "AWS - Sr Technical Account Manager",
+      "image": "https://media.licdn.com/dms/image/v2/D5603AQFQj2a90KFFZQ/profile-displayphoto-shrink_400_400/profile-displayphoto-shrink_400_400/0/1692659294509?e=1766620800&v=beta&t=wl2LxhFX4-KSVgnjn1gJtAMiDxsezjIsbgIDPAZTE1I",
+      "linkedin": "https://www.linkedin.com/in/aspagnolo/"
+    },
+    {
+      "name": "Greg Medard",
+      "bio": "AWS - Solutions Architect",
+      "image": "https://media.licdn.com/dms/image/v2/C4E03AQGveSDnRH9aCg/profile-displayphoto-shrink_400_400/profile-displayphoto-shrink_400_400/0/1565226836063?e=1762992000&v=beta&t=Jkir-cq-T1EU8MTqY7PPPxreVo2_zI4FXaXXrdTdpfQ",
+      "linkedin": "https://www.linkedin.com/in/gregorymedard/"
```
bfreiberg:
```suggestion
      "linkedin": "gregorymedard"
```

rabbitmq-private-lambda-python-sam/example-pattern.json
```diff
@@ -0,0 +1,80 @@
...
+      "name": "Kunal Ghosh",
+      "bio": "AWS - Sr SA, Strategic Accounts",
+      "image": "https://media.licdn.com/dms/image/v2/C5603AQHrj7mHd7Z1hg/profile-displayphoto-shrink_800_800/profile-displayphoto-shrink_800_800/0/1516355158121?e=1762992000&v=beta&t=8h0cpHLR6v4-e3BJ7n9Wd_OHL2rRDa8F_74rTUnu8Js",
+      "linkedin": "https://www.linkedin.com/in/kunal-ghosh-6583058/"
+    },
+    {
+      "name": "Angelo Spagnolo",
+      "bio": "AWS - Sr Technical Account Manager",
+      "image": "https://media.licdn.com/dms/image/v2/D5603AQFQj2a90KFFZQ/profile-displayphoto-shrink_400_400/profile-displayphoto-shrink_400_400/0/1692659294509?e=1766620800&v=beta&t=wl2LxhFX4-KSVgnjn1gJtAMiDxsezjIsbgIDPAZTE1I",
+      "linkedin": "https://www.linkedin.com/in/aspagnolo/"
```
bfreiberg:
```suggestion
      "linkedin": "aspagnolo"
```

rabbitmq-private-lambda-python-sam/example-pattern.json
```diff
@@ -0,0 +1,80 @@
...
+      "name": "Arghya Banerjee",
+      "bio": "AWS - Senior Solutions Architect",
+      "image": "https://media.licdn.com/dms/image/v2/C5603AQHFFKivT-1iKA/profile-displayphoto-shrink_800_800/profile-displayphoto-shrink_800_800/0/1516305772138?e=1762992000&v=beta&t=dKgoxEfgZd3M5XW-GEJ9Ss4I5ka8-C7iEIy8Sb_PTOQ",
+      "linkedin": "https://www.linkedin.com/in/arghya-b-6130b57/"
+    },
+    {
+      "name": "Kunal Ghosh",
+      "bio": "AWS - Sr SA, Strategic Accounts",
+      "image": "https://media.licdn.com/dms/image/v2/C5603AQHrj7mHd7Z1hg/profile-displayphoto-shrink_800_800/profile-displayphoto-shrink_800_800/0/1516355158121?e=1762992000&v=beta&t=8h0cpHLR6v4-e3BJ7n9Wd_OHL2rRDa8F_74rTUnu8Js",
+      "linkedin": "https://www.linkedin.com/in/kunal-ghosh-6583058/"
```
bfreiberg:
```suggestion
      "linkedin": "kunal-ghosh-6583058"
```

rabbitmq-private-lambda-python-sam/example-pattern.json
```diff
@@ -0,0 +1,80 @@
...
+      "name": "Indranil Banerjee",
+      "bio": "AWS - Senior Solutions Architect",
+      "image": "https://media.licdn.com/dms/image/v2/C5603AQEL3BG6JZca6A/profile-displayphoto-shrink_400_400/profile-displayphoto-shrink_400_400/0/1653972622784?e=1762992000&v=beta&t=a9gnmtxyWBhfEfqDF3HiPHWDoc4KZjG2sdNsIakcJXw",
+      "linkedin": "https://www.linkedin.com/in/indranil-banerjee-b00a261/"
+    },
+    {
+      "name": "Arghya Banerjee",
+      "bio": "AWS - Senior Solutions Architect",
+      "image": "https://media.licdn.com/dms/image/v2/C5603AQHFFKivT-1iKA/profile-displayphoto-shrink_800_800/profile-displayphoto-shrink_800_800/0/1516305772138?e=1762992000&v=beta&t=dKgoxEfgZd3M5XW-GEJ9Ss4I5ka8-C7iEIy8Sb_PTOQ",
+      "linkedin": "https://www.linkedin.com/in/arghya-b-6130b57/"
```
bfreiberg:
```suggestion
      "linkedin": "arghya-b-6130b57"
```

rabbitmq-private-lambda-python-sam/example-pattern.json
```diff
@@ -0,0 +1,80 @@
...
+  },
+  "cleanup": {
+    "text": ["Delete the template: <code>sam delete</code>."]
+  },
+  "authors": [
+    {
+      "name": "Indranil Banerjee",
+      "bio": "AWS - Senior Solutions Architect",
+      "image": "https://media.licdn.com/dms/image/v2/C5603AQEL3BG6JZca6A/profile-displayphoto-shrink_400_400/profile-displayphoto-shrink_400_400/0/1653972622784?e=1762992000&v=beta&t=a9gnmtxyWBhfEfqDF3HiPHWDoc4KZjG2sdNsIakcJXw",
+      "linkedin": "https://www.linkedin.com/in/indranil-banerjee-b00a261/"
```
bfreiberg:
```suggestion
      "linkedin": "indranil-banerjee-b00a261"
```

bfreiberg (changes requested):
Almost ready, few minor comments

rabbitmq-private-lambda-python-sam/RabbitMQAndClientEC2.yaml
gmedard: Good catch. Removed the rule entirely — with the VPC on 10.2.0.0/16 it matched no hosts. Port 22 access to the client is already
  covered by the self-referencing RabbitMQClientSelfIngressAllowRule and the EC2 Instance Connect CIDR that gets added at runtime in
  UserData, so nothing is lost.

rabbitmq-private-lambda-python-sam/RabbitMQAndClientEC2.yaml
gmedard: Agreed, no reason to pull from a third party. Switched to the AWS-owned https://ip-ranges.amazonaws.com/ip-ranges.json. Since
  that's the combined file rather than the pre-split per-service one, I updated the jq filter to select on
  .service=="EC2_INSTANCE_CONNECT" in addition to region.

rabbitmq-private-lambda-python-sam/README.md
gmedard: Done — updated to 2026.

rabbitmq-private-lambda-python-sam/example-pattern.json
gmedard: Done — changed to indranil-banerjee-b00a261.

rabbitmq-private-lambda-python-sam/example-pattern.json
gmedard: Done — changed to arghya-b-6130b57

rabbitmq-private-lambda-python-sam/example-pattern.json
gmedard: Done — changed to kunal-ghosh-6583058

rabbitmq-private-lambda-python-sam/example-pattern.json
gmedard: Done — changed to aspagnolo

rabbitmq-private-lambda-python-sam/example-pattern.json
gmedard: Done — changed to gregorymedard

bfreiberg:
Looks good apart from the minor comment above.

PR #3262, opened 2026-08-04
New serverless pattern - bedrock-semantic-cache-s3vectors-sam

manishh-13:
## New serverless pattern: bedrock-semantic-cache-s3vectors-sam

**Serverless semantic cache for Amazon Bedrock using AWS Lambda and Amazon S3 Vectors.**

An AWS Lambda function sits in front of Amazon Bedrock and returns a cached answer whenever an incoming prompt is *semantically* similar to one already answered, skipping the LLM call on repeats and near-repeats.

### How it works
- Lambda embeds the prompt with Amazon Titan Text Embeddings v2 (1024-dim).
- Queries an Amazon S3 Vectors index for the nearest prior prompt (cosine similarity); the cached answer is stored in the vector metadata, so no separate database is needed.
- **HIT** (similarity >= configurable threshold, within TTL, current epoch, passes negation-parity guard): returns the stored answer in ms at $0 LLM cost.
- **MISS**: calls the Bedrock text model, stores embedding + answer + metadata, returns the fresh result.

### Correctness features
- Tunable similarity threshold (precision vs. hit rate).
- Freshness TTL to bound staleness.
- One-call force-invalidation via an epoch counter in SSM Parameter Store (no scan/delete).
- Negation-parity guard (prevents "is X" vs "is NOT X" false hits).
- Top-K retrieval so a stale near-neighbour never blocks a valid hit.

### Notes
- Fully serverless: S3 Vectors scales to zero; Lambda is stateless (cache survives cold starts/redeploys).
- Access via IAM-signed Lambda function URL + optional app-level API key. Least-privilege IAM.
- Model-agnostic: the answer model is a drop-in for any Amazon Bedrock model (Amazon Nova, Anthropic Claude, Meta Llama, Mistral) via the converse API; the embedding model is configurable as long as the index dimension matches.
- Complements (does not replace) Bedrock native prompt caching, which is exact-prefix based.

Includes README.md, template.yaml (AWS SAM), example-pattern.json, and src/ (Python). Language: Python. Level: 300. Framework: AWS SAM.

Pattern-submission issue: #3263


manishh-13:
Hi, checking in on this one. It has been about four weeks with no review yet, so I wanted to make sure it did not slip through. The branch is still clean with no conflicts. Happy to make any changes you suggest. Thanks.

bedrock-semantic-cache-s3vectors-sam/template.yaml
```diff
@@ -0,0 +1,101 @@
...
+    Description: Cosine-similarity threshold for a cache hit (0-1). Higher = stricter matching.
+  TtlSeconds:
+    Type: String
+    Default: "86400"
+    Description: Freshness window in seconds. Entries older than this are treated as a miss.
+  ApiKey:
+    Type: String
+    Default: ""
+    NoEcho: true
+    Description: Optional app-level API key checked in the x-api-key header. Leave blank to rely only on IAM auth.
```
marcojahn, line 29: The optional application-level API key is passed via a `NoEcho` parameter (good) but is then injected into the Lambda function's `API_KEY` environment variable in plaintext. AWS explicitly recommends against storing secrets in Lambda environment variables because they are visible to anyone with permission to view the function configuration. Lambda encrypts environment variables at rest with an AWS-managed key by default, but the value is still readable in the console/API.

**Recommended Fix:**
  ```yaml
  # Store the key in SSM Parameter Store as a SecureString and resolve it at
  # deploy time, or fetch it at runtime. Example using dynamic reference:
      Environment:
        Variables:
          API_KEY_PARAM: /semantic-cache/api-key   # SecureString parameter
  # Grant ssm:GetParameter + kms:Decrypt on that parameter, and read it in
  # cache.py at cold start via ssm.get_parameter(..., WithDecryption=True).
  # Alternatively use AWS Secrets Manager with the Parameters and Secrets
  # Lambda extension for cached retrieval.
  ```

bedrock-semantic-cache-s3vectors-sam/template.yaml
```diff
@@ -0,0 +1,101 @@
...
+              Action: bedrock:InvokeModel
+              Resource: "arn:aws:bedrock:*::foundation-model/*"
+            - Sid: S3Vectors
+              Effect: Allow
+              Action:
+                - s3vectors:QueryVectors
+                - s3vectors:GetVectors
+                - s3vectors:ListVectors
+                - s3vectors:PutVectors
+                - s3vectors:GetIndex
```
marcojahn, line 84: The policy grants `s3vectors:ListVectors` and `s3vectors:GetIndex`, but `cache.py` only calls `query_vectors` and `put_vectors`. `QueryVectors` with `returnMetadata=true` additionally requires `GetVectors` (correctly granted). Removing the two unused actions tightens least privilege further.


**Recommended Fix:**
  ```yaml
  Action:
    - s3vectors:QueryVectors
    - s3vectors:GetVectors   # required because query_vectors uses returnMetadata=True
    - s3vectors:PutVectors
  ```

bedrock-semantic-cache-s3vectors-sam/README.md
```diff
@@ -0,0 +1,151 @@
...
+
+- An AWS account with permissions for AWS Lambda, Amazon Bedrock, Amazon S3 Vectors, and AWS Systems Manager.
+- [AWS CLI](https://docs.aws.amazon.com/cli/latest/userguide/install-cliv2.html) v2, recent enough to include the `s3vectors` commands.
+- [AWS SAM CLI](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/serverless-sam-cli-install.html).
+- Amazon Bedrock **model access enabled** for the embeddings model (`amazon.titan-embed-text-v2:0`) and the text model (`amazon.nova-lite-v1:0`) in your Region.
+- A Region where Amazon S3 Vectors and Amazon Bedrock are available (e.g. `us-east-1`).
+
+## Deployment
+
+S3 Vectors is not yet a CloudFormation resource, so create the vector store first (two commands), then deploy the rest with SAM.
```
marcojahn, line 81: Amazon S3 Vectors now has CloudFormation support, so this statement is inaccurate and steers users toward manual CLI steps that could instead be codified in the template.

```markdown
  Amazon S3 Vectors resources (AWS::S3Vectors::VectorBucket and
  AWS::S3Vectors::Index) can now be defined directly in the SAM/CloudFormation
  template, removing the manual setup commands. If you prefer to keep the
  vector store decoupled from the stack lifecycle, the two CLI commands below
  remain a valid alternative.
  ```

File: bedrock-semantic-cache-s3vectors-sam/README.md [hunk 4]
```diff
@@ -0,0 +1,151 @@
...
+
+```bash
+# 1. Create the S3 Vectors bucket and a cosine index (1024 dims = Titan v2)
+export VECTOR_BUCKET="semantic-cache-$(aws sts get-caller-identity --query Account --output text)"
```
marcojahn, line 85: Deployment jumps straight to the `aws s3vectors` and `sam build` commands without a step to obtain the code and `cd` into the pattern directory. Most users arrive via the serverlessland link, but an explicit clone/cd step makes the instructions self-contained.

```markdown
  git clone https://github.com/aws-samples/serverless-patterns
  cd serverless-patterns/bedrock-semantic-cache-s3vectors-sam
  # then continue with the S3 Vectors setup and sam build/deploy below
  ```

File: bedrock-semantic-cache-s3vectors-sam/README.md [hunk 5]
```diff
@@ -0,0 +1,151 @@
...
+- **top-K + iterate** - a stale duplicate near-neighbour never blocks a valid hit.
+
+## Requirements
+
+- An AWS account with permissions for AWS Lambda, Amazon Bedrock, Amazon S3 Vectors, and AWS Systems Manager.
+- [AWS CLI](https://docs.aws.amazon.com/cli/latest/userguide/install-cliv2.html) v2, recent enough to include the `s3vectors` commands.
+- [AWS SAM CLI](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/serverless-sam-cli-install.html).
+- Amazon Bedrock **model access enabled** for the embeddings model (`amazon.titan-embed-text-v2:0`) and the text model (`amazon.nova-lite-v1:0`) in your Region.
+- A Region where Amazon S3 Vectors and Amazon Bedrock are available (e.g. `us-east-1`).
+
```
marcojahn, line 78: The Testing section uses `python3 -c "..."` to build invoke payloads, but Python is not listed in the Requirements section. SAM builds the function runtime independently, so Python is not needed to deploy — only to run the documented test snippets. Noting it prevents confusion.

```markdown
  ## Requirements
  * ... (existing items)
  * Python 3 (only needed to run the test payload helper commands below)
  ```

bedrock-semantic-cache-s3vectors-sam/README.md
marcojahn, line 1: The README introduces the full name "AWS Systems Manager" but later uses the abbreviation "SSM Parameter Store". For consistency, prefer "AWS Systems Manager Parameter Store" (or "Systems Manager" after first reference).

marcojahn (commented):
Hello @manishh-13, I've reviewed the pattern and created a few findings. Please review.

manishh-13:
Thanks for the review @marcojahn. All six points are addressed in the two commits I just pushed. Details below.

**1. API key in a plaintext Lambda environment variable (template.yaml:29)**
Removed the `ApiKey` NoEcho parameter and the `API_KEY` environment variable. The template now takes an optional `ApiKeyParameterName` pointing at an existing AWS Systems Manager Parameter Store SecureString parameter, and only that parameter *name* is passed to the function. The function reads the value at runtime with `get_parameter(..., WithDecryption=True)` and caches it per execution environment, so the secret never appears in the function configuration. The two matching IAM statements (`ssm:GetParameter` on that one parameter, `kms:Decrypt` scoped with `kms:ViaService: ssm.<region>.amazonaws.com`) are added conditionally and only when the parameter name is supplied. Leaving it blank keeps the function on IAM auth alone.

**2. Unused S3 Vectors IAM actions (template.yaml:84)**
Dropped `s3vectors:ListVectors` and `s3vectors:GetIndex`. The policy is now exactly the three actions the code calls: `QueryVectors`, `GetVectors` and `PutVectors`. `GetVectors` is kept because `query_vectors` is called with `returnMetadata=True`, which requires it.

**3. "S3 Vectors is not yet a CloudFormation resource" (README.md:81)**
Correct, that is out of date. `AWS::S3Vectors::VectorBucket` and `AWS::S3Vectors::Index` both exist now, so the claim and the manual `aws s3vectors create-*` steps are gone. The template creates both resources, the index is configured to match the embeddings model (dimension 1024, cosine, float32) and the function has a `DependsOn` on the index. Cleanup is now just `sam delete`.

**4. Missing clone step (README.md:85)**
Added the `git clone` and `cd` steps at the top of the Deployment section.

**5. Python not listed in Requirements (README.md:78)**
Added Python 3 to Requirements, since the test commands use `python3`.

**6. "SSM" vs the full service name (README.md:1)**
Normalized to "AWS Systems Manager Parameter Store" on first use throughout, with no bare "SSM" left in the prose.

One extra fix while I was in there: the epoch parameter name and the Lambda function name were hardcoded, which made a second deployment in the same account fail early validation. Both are now derived from the stack name.

Re-verified end to end in a clean account after the changes: cache miss on the first prompt, hit on a repeat (similarity 0.9996, 221 ms), hit on a reworded prompt (similarity 0.9562), 401 on a wrong `x-api-key` (which confirms the Parameter Store read and the scoped KMS grant work), and the invalidate action bumps the epoch. `sam validate --lint` is clean.

PR #3308, opened 2026-09-17
smk-lambda-queue-mode-python-sam: Self-managed Kafka Queue mode (KIP-932) to Lambda

hardith:
## Description
Adds a new pattern that deploys an AWS Lambda function consuming from a self-managed Apache Kafka 4.2+ cluster using **Queue consumption mode** (KIP-932 Share Groups). In Queue mode, multiple Lambda pollers can process records from the same partition concurrently — parallelism is decoupled from partition count, unlike Stream mode where each partition maps to a single consumer.

## Architecture
Producer Lambda -> self-managed Apache Kafka 4.2+ cluster (KRaft, share groups enabled) -> Lambda ESM in Queue mode (`ConsumptionMode: Queue`, provisioned pollers 2-10) -> Worker Lambda (Python 3.12, partial batch response) -> Amazon SQS DLQ for records that exhaust `MaximumRetryAttempts`. Observability via a CloudWatch dashboard and alarms for share-group lag, DLQ delivery, and poller errors.

Deployment is split into four independent stacks (network, broker, app, observability) so users can bring their own Kafka cluster and/or VPC.

## Language / Framework
Python 3.12, AWS SAM + AWS CloudFormation.

## Note for reviewers — `ConsumptionMode: Queue` schema
`ConsumptionMode: Queue` is a pre-release Lambda API field not yet in the AWS CLI or CloudFormation schema. The pattern therefore creates the ESM via `scripts/create-esm.sh`, which calls the Lambda REST API directly using `curl --aws-sigv4`. AWS CLI support for this field is landing on **Sep 25, 2026**. Once available, we intend to submit a follow-up PR replacing the curl call in `create-esm.sh` with the standard `aws lambda create-event-source-mapping` CLI command and updating the accompanying blog.

## Testing
Deployed and validated end-to-end (produce, consume, retry, DLQ, share-group behavior with concurrent pollers > partitions).


smk-lambda-queue-mode-python-sam/stacks/3-app.yaml
```diff
@@ -0,0 +1,143 @@
...
+              Resource: !GetAtt TaskWorkerDLQ.Arn
+            - Effect: Allow
+              Action:
+                - ec2:CreateNetworkInterface
+                - ec2:DescribeNetworkInterfaces
+                - ec2:DeleteNetworkInterface
+                - ec2:DescribeVpcs
+                - ec2:DescribeSubnets
+                - ec2:DescribeSecurityGroups
+              Resource: '*'
```
bfreiberg: `WorkerFunction` is granted `ec2:CreateNetworkInterface`, `DescribeNetworkInterfaces`, `DeleteNetworkInterface`, etc., but the function has no `VpcConfig`. The worker is invoked by the ESM pollers and runs in the Lambda-managed VPC — it never creates ENIs in the customer VPC (the pollers handle Kafka connectivity via `SourceAccessConfigurations`). These permissions appear unused and can be removed; the inline policy can be reduced to just the `sqs:SendMessage` statement on the DLQ.

smk-lambda-queue-mode-python-sam/README.md
```diff
@@ -0,0 +1,374 @@
...
+chmod +x scripts/setup-broker.sh
+./scripts/setup-broker.sh --region <region> --profile <profile>
+```
+
+This script connects to the broker via SSM (no SSH required) and installs Apache Kafka 4.2.x, configures KRaft mode with share groups enabled, and starts the broker. It runs 6 steps sequentially and reports progress. The Kafka download (~130MB) takes about 15-20 minutes depending on network speed.
+
+**Step 3: Build and deploy the application**
+
+```bash
+sam build --template stacks/3-app.yaml
```
bfreiberg: `confluent-kafka` ships as a compiled wheel, so a plain `sam build` on a host whose platform/Python doesn't match the Lambda `python3.12` x86_64 runtime can produce an incompatible artifact that fails at import time. Recommend `sam build --template stacks/3-app.yaml --use-container` to guarantee a Lambda-compatible build (verified working). Worth applying to the Path B block too.

smk-lambda-queue-mode-python-sam/scripts/create-esm.sh
```diff
@@ -0,0 +1,139 @@
...
+echo "Exporting credentials for SigV4 signing..."
+unset AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY AWS_SESSION_TOKEN
+eval "$(aws configure export-credentials --format env --profile "$PROFILE")"
+
+# ── Create ESM via Lambda REST API ───────────────────────────
+echo "Creating ESM..."
+RESPONSE=$(curl -sS -X POST \
+  "https://lambda.$REGION.amazonaws.com/2015-03-31/event-source-mappings/" \
+  --aws-sigv4 "aws:amz:$REGION:lambda" \
+  --user "$AWS_ACCESS_KEY_ID:$AWS_SECRET_ACCESS_KEY" \
```
bfreiberg: `curl --user "$AWS_ACCESS_KEY_ID:$AWS_SECRET_ACCESS_KEY"` puts the secret access key in the process argument list, where it's visible via `ps` for the duration of the call. It's short-lived and this is a sample, so minor — but a one-line note in the script header (or moving to a `~/.netrc`/stdin approach) would be a good hygiene callout, especially since folks copy sample scripts.

smk-lambda-queue-mode-python-sam/stacks/3-app.yaml
bfreiberg: `WorkerFunction` is granted `ec2:CreateNetworkInterface`, `DescribeNetworkInterfaces`, `DeleteNetworkInterface`, etc., but the function has no `VpcConfig`. The worker is invoked by the ESM pollers and runs in the Lambda-managed VPC — it never creates ENIs in the customer VPC (the pollers handle Kafka connectivity via `SourceAccessConfigurations`). These permissions appear unused and can be removed; the inline policy can be reduced to just the `sqs:SendMessage` statement on the DLQ.

smk-lambda-queue-mode-python-sam/README.md
bfreiberg: `confluent-kafka` ships as a compiled wheel (wraps librdkafka), so a plain `sam build` on a host whose platform/Python doesn't match the Lambda `python3.12` x86_64 runtime can produce an incompatible artifact that fails at import time. Recommend `sam build --template stacks/3-app.yaml --use-container` to guarantee a Lambda-compatible build (verified working). Worth applying to the Path B block too.

smk-lambda-queue-mode-python-sam/scripts/create-esm.sh
bfreiberg, line 85: `curl --user "$AWS_ACCESS_KEY_ID:$AWS_SECRET_ACCESS_KEY"` puts the secret access key in the process argument list, where it's visible via `ps` for the duration of the call. It's short-lived and this is a sample, so minor — but a one-line note in the script header (or moving to a `~/.netrc`/stdin approach) would be a good hygiene callout, especially since folks copy sample scripts.

smk-lambda-queue-mode-python-sam/stacks/2-broker.yaml
```diff
@@ -0,0 +1,55 @@
...
+    Properties:
+      Roles:
+        - !Ref KafkaInstanceRole
+
+  KafkaInstance:
+    Type: AWS::EC2::Instance
+    Properties:
+      InstanceType: !Ref InstanceType
+      ImageId: !Ref LatestAL2023AmiId
+      SubnetId: !ImportValue kafka-queue-PublicSubnetId
```
bfreiberg, line 38: **[MEDIUM · security]** The broker EC2 instance is launched into `PublicSubnet`, which sets `MapPublicIpOnLaunch: true` (`1-network.yaml:35`) and is routed `0.0.0.0/0` to the Internet Gateway — so the broker gets a public IP and is internet-facing. Placing a data-plane broker in a public subnet is the same anti-pattern as a public-subnet datastore.

smk-lambda-queue-mode-python-sam/src/producer/producer.py
```diff
@@ -0,0 +1,74 @@
...
+logger = logging.getLogger()
+logger.setLevel(logging.INFO)
+
+TOPIC = os.environ.get("KAFKA_TOPIC", "kafka-queue-task-worker")
+BOOTSTRAP_SERVERS = os.environ.get("BOOTSTRAP_SERVERS", "")
+DEFAULT_COUNT = 50
+
+
+def _config():
+    return {"bootstrap.servers": BOOTSTRAP_SERVERS}
```
bfreiberg, line 24: **[MEDIUM · security]** The end-to-end data path is PLAINTEXT with no auth: broker listener `PLAINTEXT://0.0.0.0:9092`, the producer connects with only `bootstrap.servers` (no TLS/SASL) here, and the ESM uses a plain VPC source config. Fine for an isolated demo, but the README body should prominently call out that traffic is unencrypted/unauthenticated and point to SASL/SCRAM + TLS for any non-sandbox use.

smk-lambda-queue-mode-python-sam/stacks/1-network.yaml
```diff
@@ -0,0 +1,196 @@
...
+    Type: AWS::EC2::SecurityGroupIngress
+    Properties:
+      GroupId: !Ref LambdaSecurityGroup
+      IpProtocol: tcp
+      FromPort: 443
+      ToPort: 443
+      SourceSecurityGroupId: !Ref LambdaSecurityGroup
+      Description: Self - Hyperplane ENI to VPC endpoint ENI
+
+  # VPC Interface Endpoints — all three are required for Queue mode ESM pollers:
```
bfreiberg, line 149: **[MEDIUM · architecture]** the lambda/sts/sqs VPC endpoints (`setup-vpc-endpoints.sh`), the broker install (`setup-broker.sh`), and the ESM itself (`create-esm.sh`) live in shell scripts rather than the CloudFormation/SAM stacks. That splits the deployable surface and makes the runtime wiring non-reproducible and not drift-detectable. The ESM is understandably scripted while `ConsumptionMode: Queue` is preview-only, but the VPC endpoints are plain CloudFormation resources and could move into this stack.

smk-lambda-queue-mode-python-sam/src/worker/handler.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def lambda_handler(event, context):
+    failures = []
+
+    for tp_key, records in event.get("records", {}).items():
+        for r in records:
+            try:
+                payload = json.loads(base64.b64decode(r["value"]).decode("utf-8"))
+            except Exception:
```
bfreiberg, line 19: **[MEDIUM · code-quality]** This bare `except Exception` silently replaces a malformed/non-JSON record with a raw-payload fallback and logs nothing. Since this sample is copied verbatim into learner accounts, a real base64/JSON decode bug (e.g. an unexpected producer payload) would be masked. Recommend logging before the fallback, e.g. `logger.warning("failed to decode record %s-%s-%s", r.get('topic'), r.get('partition'), r.get('offset'), exc_info=True)`.

smk-lambda-queue-mode-python-sam/README.md
```diff
@@ -0,0 +1,374 @@
...
+
+- [AWS CLI](https://docs.aws.amazon.com/cli/latest/userguide/install-cliv2.html) installed and configured
+- [AWS SAM CLI](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/serverless-sam-cli-install.html) installed
+- [Python 3.12](https://www.python.org/downloads/)
+- curl >= 7.75 (for `--aws-sigv4` support)
+- Apache Kafka 4.2+ cluster (see Deployment Path A to provision one automatically)
+
+## Costs
+
+This pattern uses EC2 (t3.medium), Lambda, SQS, CloudWatch, and VPC resources. See [AWS Pricing](https://aws.amazon.com/pricing/) for details. There are costs associated with these services beyond the Free Tier.
```
bfreiberg: **[MEDIUM · naming]** First prose mention of these services should use the full official names: **Amazon EC2**, **Amazon SQS**, **Amazon CloudWatch**, and **Amazon VPC** (`Lambda` here is fine — `AWS Lambda` was already introduced above). Serverless Land style expects the full service name on first reference.

smk-lambda-queue-mode-python-sam/example-pattern.json
```diff
@@ -0,0 +1,70 @@
+{
+  "title": "Self-managed Apache Kafka Queue mode to AWS Lambda (KIP-932)",
+  "description": "Deploy a Lambda function that consumes from a self-managed Apache Kafka 4.2+ cluster using Queue consumption mode (KIP-932 Share Groups). Queue mode allows multiple Lambda pollers to process records from the same partition concurrently, breaking the partition-count ceiling of traditional consumer groups.",
+  "language": "Python",
+  "level": "300",
+  "framework": "AWS SAM",
```
bfreiberg: **[MEDIUM · schema]** `framework` is `"AWS SAM"`; the recognized enum value is `"SAM"`. If the serverlessland.com listing filter matches the exact enum string, this pattern may not group with other SAM patterns. The IaC itself is correct (`Transform: AWS::Serverless-2016-10-31`). Worth confirming the accepted value against the current metadata schema / a recently-merged SAM pattern.

smk-lambda-queue-mode-python-sam/example-pattern.json
```diff
@@ -0,0 +1,70 @@
...
+  },
+  "introBox": {
+    "headline": "How it works",
+    "text": [
+      "With Queue consumption mode (ConsumptionMode: Queue), Lambda event pollers use Kafka Share Groups (KIP-932) instead of traditional consumer groups. Multiple pollers can read from the same partition concurrently — parallelism is decoupled from partition count.",
+      "This pattern deploys: a dedicated VPC with private subnets and VPC endpoints, a self-managed Kafka 4.2+ broker on EC2 with share groups enabled, a Python Lambda worker function with partial batch response, a producer Lambda for testing, and CloudWatch observability.",
+      "Deployment is split into four independent stacks so you can skip the network and broker stacks if you have existing infrastructure."
+    ]
+  },
+  "gitHub": {
```
bfreiberg, line 19: **[MEDIUM · schema]** The `gitHub` object nests `template` as an object (`repoURL`/`templateURL`/`projectFolder`/`templateFile`) and has no direct `projectFolder`. Some versions of the metadata schema expect `gitHub.template` and `gitHub.projectFolder` as direct string sub-properties. Depending on how serverlessland.com reads this, the template link/folder may not resolve. Worth confirming this structure against a recently-merged pattern's `example-pattern.json` before merge.

smk-lambda-queue-mode-python-sam/README.md
```diff
@@ -0,0 +1,374 @@
...
+
+| Property | Value |
+|----------|-------|
+| Kafka version required | Apache Kafka 4.2+ |
+| Lambda runtime | Python 3.12 |
+| IaC framework | AWS SAM + AWS CloudFormation |
+| Delivery semantics | At-least-once |
+| Ordering guarantees | None (Queue mode) |
+| Authentication | PLAINTEXT (see notes for SASL/SCRAM) |
+| Region | Configurable |
```
bfreiberg: Remove but add copyright footer

smk-lambda-queue-mode-python-sam/example-pattern.json
```diff
@@ -0,0 +1,70 @@
+{
+  "title": "Self-managed Apache Kafka Queue mode to AWS Lambda (KIP-932)",
+  "description": "Deploy a Lambda function that consumes from a self-managed Apache Kafka 4.2+ cluster using Queue consumption mode (KIP-932 Share Groups). Queue mode allows multiple Lambda pollers to process records from the same partition concurrently, breaking the partition-count ceiling of traditional consumer groups.",
```
bfreiberg: description must be 175 characters or fewer

smk-lambda-queue-mode-python-sam/example-pattern.json
```diff
@@ -0,0 +1,70 @@
...
+  "cleanup": {
+    "text": [
+      "Delete the ESM first: <code>aws lambda delete-event-source-mapping --uuid &lt;esm-uuid&gt;</code>, then delete stacks in reverse order: <code>kafka-queue-observability</code>, <code>kafka-queue-app</code>, <code>kafka-queue-broker</code>, <code>kafka-queue-network</code>."
+    ]
+  },
+  "authors": [
+    {
+      "name": "Vaibhav Jain",
+      "bio": "AWS - Senior Delivery Consultant. Specializes in data platform modernization and large-scale distributed systems.",
+      "linkedin": "https://www.linkedin.com/in/vaibhavjainv/"
```
bfreiberg:
```suggestion
      "linkedin": "vaibhavjainv"
```

smk-lambda-queue-mode-python-sam/example-pattern.json
```diff
@@ -0,0 +1,70 @@
...
+    },
+    {
+      "name": "Hardith Murari",
+      "bio": "AWS - Delivery Consultant. Specializes in Application Architecture and Modernization.",
+      "linkedin": "https://www.linkedin.com/in/hardithsuvarna/"
```
bfreiberg:
```suggestion
      "linkedin": "hardithsuvarna"
```

smk-lambda-queue-mode-python-sam/stacks/1-network.yaml
```diff
@@ -0,0 +1,196 @@
...
+  #   sts     — pollers obtain temporary credentials
+  #   sqs     — pollers write failed records to the OnFailure DLQ
+  #
+  # IMPORTANT: Missing the SQS endpoint causes a silent connection error —
+  # the ESM stays Enabled/OK but Lambda stops being invoked after the first
+  # batch when a failed record triggers a DLQ write.
+  #
+  # These are created via scripts/setup-vpc-endpoints.sh rather than as
+  # CloudFormation resources because AWS::EC2::VpcEndpoint is restricted
+  # in some accounts (e.g. Isengard/bindled accounts used for testing).
```
bfreiberg: This leaks internal information, remove

bfreiberg (changes requested):

bfreiberg (approved):
Looks good, thanks for your contribution. Your pattern will be merged soon.

